PR #22, opened 2019-12-07
#21 Make it easier to add operation actions

### LINKED ISSUE #21: OperationActionHandler makes it hard to add custom operation action handlers
Currently, the `OperationActionHandler` handles operation actions in the following way:
```java
@Override
public Optional<Action> execute(Action action, GraphicalModelState modelState) {
	switch (action.getKind()) {
	case Action.Kind.CREATE_NODE_OPERATION:
	case Action.Kind.CREATE_CONNECTION_OPERATION:
	case Action.Kind.RECONNECT_CONNECTION_OPERATION:
	case Action.Kind.REROUTE_CONNECTION_OPERATION:
	case Action.Kind.DELETE_ELEMENT_OPERATION:
	case Action.Kind.CHANGE_BOUNDS_OPERATION:
	case Action.Kind.APPLY_LABEL_EDIT_OPERATION:
		return doHandle((AbstractOperationAction) action, modelState);
	default:
		return Optional.empty();
	}
}
```

That means, if users would like to add custom operation actions, they'll have to overwrite the `OperationActionHandler` and add their new action kind in the check above. This is unexpected and also unnecessary. We could do the following instead:

```java
@Override
public Optional<Action> execute(final Action action, final GraphicalModelState modelState)  {
  if (action instanceof AbstractOperationAction && operationHandlerProvider.isHandled((AbstractOperationAction)action)) {
      return doHandle((AbstractOperationAction)action, modelState);
  }
  return Optional.empty();
}
```

planger:
Resolves #21

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actionhandler/OperationActionHandler.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actionhandler/OperationActionHandler.java
@@ -37,20 +37,12 @@ public class OperationActionHandler extends AbstractActionHandler {
    }
 
    @Override
-   @SuppressWarnings("checkstyle:CyclomaticComplexity")
    public Optional<Action> execute(final Action action, final GraphicalModelState modelState) {
-      switch (action.getKind()) {
-         case Action.Kind.CREATE_NODE_OPERATION:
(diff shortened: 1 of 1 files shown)


tortmayr:
Has to be adpated after #23  was merged

PR #23, opened 2019-12-10
Swap to eclipse-plugin based infra and include m2-only build

tortmayr:
This enables dedicated builds for plain maven and Eclipse p2 within the same repository. I also configured corresponding deployment jobs and updated the Jenkinsfile.
The testing setup is still missing and will be added in a follow-up PR.


releng/org.eclipse.glsp.repository/test
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0"?>
```
planger: The file name looks strange. Was this a test and is a left-over or is this required and could be named differently?

tortmayr: Good catch. Yes this is a leftover and can be remvoed

PR #36, opened 2020-01-22
Ensure that JavaSE-1.8 is used instead of 10

tortmayr:
This avoids Eclipse issues with Java10.  This also aligns the plugin configuration and the maven configuration (which uses Java 1.8 anyways)


planger:
Thanks!

PR #37, opened 2020-01-22
Refactor DI multibinding approach

tortmayr:
- Introduce new helper class for easier configuration of multi bindings (add,remove,rebind)
- Remove ActionProvider and bind Action classes directly via multi bind
- Introduce InstanceRegistry (client side already has a similar concept)
- Refactor ActionRegistry to make use of the new InstanceRegistry
- Move multibinding default values into dedicated util class


tortmayr:
needs https://github.com/eclipse-glsp/glsp-examples/pull/30

PR #52, opened 2020-03-16
Fallback to model search when the GModelIndexImpl is still indexing

planger:
Fixes https://github.com/eclipse-glsp/glsp/issues/50

### DIFF
--- a/plugins/org.eclipse.glsp.graph/src/org/eclipse/glsp/graph/impl/GModelIndexImpl.java
+++ b/plugins/org.eclipse.glsp.graph/src/org/eclipse/glsp/graph/impl/GModelIndexImpl.java
@@ -30,6 +30,7 @@ import java.util.stream.Collectors;
 
 import org.eclipse.emf.common.notify.Notification;
 import org.eclipse.emf.common.notify.Notifier;
+import org.eclipse.emf.common.util.TreeIterator;
 import org.eclipse.emf.ecore.EClass;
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.glsp.graph/src/org/eclipse/glsp/graph/impl/GModelIndexImpl.java
```diff
@@ -118,7 +119,11 @@ protected void handleContainment(final Notification notification) {
 
    @Override
    public Optional<GModelElement> get(final String elementId) {
-      return Optional.ofNullable(idToElement.get(elementId));
+      Optional<GModelElement> indexMatch = Optional.ofNullable(idToElement.get(elementId));
+      if (indexMatch.isPresent() || !isCurrentlyBuildingIndex()) {
```
martin-fleck-at: My brain has a hard time parsing OR... NOT, I find the other way usually easier, e.g., isEmpty AND isCurrentlyBuilding..., but that is probably just me ;-)

planger: Yeah, I agree... I've now changed it to `!isPresent && isCurrentlyBuildingIndex()`. We can't use isEmpty because of the Java-8 compatibility. But still, I think it is better this way. Thanks!

PR #66, opened 2020-07-29
Fixes #21: Align action processor with client-side action dispatcher

### LINKED ISSUE #21: OperationActionHandler makes it hard to add custom operation action handlers

martin-fleck-at:
- Action processor should invoke ALL registered action handlers
- Each response by an action handler should be processed
- Introduce client action handler that sends actions to client

Note: This change does not adapt the operation handlers which execute operations and do not return any actions as of now.

Signed-off-by: Martin Fleck <mfleck@eclipsesource.com>


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/di/MultiBindingDefaults.java
```diff
@@ -88,7 +97,22 @@ private MultiBindingDefaults() {}
       TriggerEdgeCreationAction.class,
       UpdateModelAction.class);
 
+   public static final List<Class<? extends Action>> DEFAULT_CLIENT_ACTIONS = Lists.newArrayList(
+      SetClipboardDataAction.class,
```
tortmayr, line 101: It looks like this list is not complete. There are a couple of other client actions (ike. Center, FitToScreen, Select, Expand, ServerStatusAction etc.)

We should also double check the ActionHandlers e.g. the `SetEditModeActionHandler` implicitly registers the `SetEditModeAction` which has to be registered as client action as well.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/action/DefaultActionProcessor.java
```diff
@@ -15,23 +15,26 @@
...
 import org.eclipse.glsp.api.jsonrpc.GLSPClientProvider;
 import org.eclipse.glsp.api.model.GraphicalModelState;
 import org.eclipse.glsp.api.model.ModelStateProvider;
 import org.eclipse.glsp.api.registry.ActionHandlerRegistry;
+import org.eclipse.glsp.server.actionhandler.RequestEditValidationHandler;
 
 import com.google.inject.Inject;
 
 public class DefaultActionProcessor implements ActionProcessor {
+   private static Logger LOG = Logger.getLogger(RequestEditValidationHandler.class);
```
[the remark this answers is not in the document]
martin-fleck-at: Fixed ;-)

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/di/MultiBindingDefaults.java
martin-fleck-at: Very good point! As discussed, I registered the necessary actions as client actions and also removed up some unused actions, cf also https://github.com/eclipse-glsp/glsp-client/pull/74.

I also added a logging example for the server-side action dispatching, cf. https://github.com/eclipse-glsp/glsp-examples/pull/60

martin-fleck-at: Very good observation! I tested this and you are right, the DIActionRegistry already captures all handled actions through the action and operation handler bindings. As a result, the default action binding was removed altogether. I pushed a new commit to reflect that. Thank you!

PR #69, opened 2020-08-12
#95 Cleanup GLSPServerLauncher & add CLI parsing

tortmayr:
- Cleanup GLSPServerLauncher (refactor methods & add support to inject additional modules)
- Add a CLIParser helper class and configure default options for configuring
- - Server port
- - consoleLog flag
- - logging directory
- - logLevel
- - help

Fixes eclipse-glsp/glsp/issues/95

### DIFF
--- a/plugins/org.eclipse.glsp.server/META-INF/MANIFEST.MF
+++ b/plugins/org.eclipse.glsp.server/META-INF/MANIFEST.MF
@@ -12,7 +12,8 @@ Require-Bundle: org.eclipse.glsp.api;bundle-version="0.7.0";visibility:=reexport
  com.google.gson;bundle-version="2.8.2",
  org.eclipse.lsp4j;bundle-version="0.7.0",
  org.eclipse.lsp4j.jsonrpc,
- com.google.guava;bundle-version="21.0.0"
+ com.google.guava;bundle-version="21.0.0",
+ org.apache.commons.cli;bundle-version="1.4.0";visibility:=reexport
 Import-Package: com.google.inject.multibindings;version="1.3.0"
 Export-Package: org.eclipse.glsp.server.actionhandler,
  org.eclipse.glsp.server.command,
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/launch/CLIParser.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/launch/CLIParser.java
@@ -15,26 +15,22 @@
  ********************************************************************************/
 package org.eclipse.glsp.server.launch;
 
-import java.io.File;
+import java.util.function.Predicate;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.DefaultParser;
-import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
-import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
 import org.eclipse.glsp.server.utils.LaunchUtil;
 
-public final class CLIParser {
+public abstract class CLIParser {
    private static final Logger LOG = Logger.getLogger(CLIParser.class);
-   private final CommandLine cmd;
-   private final Options options;
-   private final String processName;
+   protected static final String INVALID_ARGUMENT_MESSAGE = "%s' is not a valid argument for option '--%s'! The default value '%s' is used.";
 
-   public CLIParser(final String[] args, final String processName) throws ParseException {
-      this(args, getDefaultCLIOptions(), processName);
-   }
+   protected final CommandLine cmd;
+   protected final Options options;
+   protected final String processName;
 
    public CLIParser(final String[] args, final Options options, final String processName) throws ParseException {
       this.cmd = new DefaultParser().parse(options, args);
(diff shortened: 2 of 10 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/launch/CLIParser.java
```diff
@@ -0,0 +1,108 @@
...
+      return logDir;
+   }
+
+   public Level parseLogLevel() {
+      String levelArg = cmd.getOptionValue("ll");
+      return Level.toLevel(levelArg, LaunchUtil.DEFAULT_LOG_LEVEL);
+   }
+
+   public boolean isConsoleLog() { return optionExists("c"); }
+
```
martin-fleck-at, line 88: Since we have utility methods for almost all parameters I suggest also adding one for 'isHelp' that evaluates the 'h'-option.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/launch/CLIParser.java
```diff
@@ -0,0 +1,108 @@
...
+      this(args, getDefaultCLIOptions(), processName);
+   }
+
+   public CLIParser(final String[] args, final Options options, final String processName) throws ParseException {
+      this.cmd = new DefaultParser().parse(options, args);
+      this.options = options;
+      this.processName = processName;
+   }
+
+   public boolean optionExists(final String identifier) {
```
martin-fleck-at, line 45: Just out of curiosity: Is there a reason you chose 'optionExists' over 'hasOption'? In my opinion 'hasOption' is both shorter and more common.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/launch/CLIParser.java
```diff
@@ -0,0 +1,108 @@
...
+      return Level.toLevel(levelArg, LaunchUtil.DEFAULT_LOG_LEVEL);
+   }
+
+   public boolean isConsoleLog() { return optionExists("c"); }
+
+   public void printHelp() {
+      printHelp(this.processName, options);
+   }
+
+   public static void printHelp(final String processName, final Options options) {
```
martin-fleck-at, line 93: Super nit-picky: It seems to me that this method would be better located at the LaunchUtil class as it has no connection to the CLIParser whatsoever besides that it takes options.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/utils/LaunchUtil.java
```diff
@@ -0,0 +1,81 @@
...
+      if (!root.getAllAppenders().hasMoreElements()) {
+         root.addAppender(new ConsoleAppender(
+            new PatternLayout(PatternLayout.TTCC_CONVERSION_PATTERN)));
+      }
+      root.setLevel(logLevel);
+
+   }
+
+   public static void configureLogger(
+      final CLIParser parser) throws ParseException, IOException {
```
martin-fleck-at, line 57: Unnecessary linebreak.

tortmayr:
Thanks for the fast review. I simply mostly reused the CLIParser class from https://github.com/eclipse-emfcloud/emfcloud-modelserver. So we should probably incorporate the mentioned improvements there as well.

plugins/org.eclipse.glsp.server/META-INF/MANIFEST.MF
```diff
@@ -12,7 +12,8 @@ Require-Bundle: org.eclipse.glsp.api;bundle-version="0.7.0";visibility:=reexport
  com.google.gson;bundle-version="2.8.2",
  org.eclipse.lsp4j;bundle-version="0.7.0",
  org.eclipse.lsp4j.jsonrpc,
- com.google.guava;bundle-version="21.0.0"
+ com.google.guava;bundle-version="21.0.0",
+ org.apache.commons.cli;bundle-version="1.4.0";visibility:=reexport
```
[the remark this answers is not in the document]
tortmayr: In this case we should be on the save side. The latest release of commons.cli is from 2017 :wink:. However,  I doublechecked and we are also reexporting some dependencies  (guice, javax.websocket, ecore etc.). Maybe we should reconsider this.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/utils/LaunchUtil.java
```diff
@@ -0,0 +1,81 @@
...
+      if (parser.isConsoleLog()) {
+         configureConsoleLogger();
```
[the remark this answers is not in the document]
tortmayr: I agree that it sometimes might be useful to have both loggers configured. However, we also need to possibility to disable console logging (if the client-server connection is handled via sys in/out streams).

I suggest the following: 
- Console logging is enabled per default
- File logging can be enabled/disabled via option
- Console logging can be enabled/disabled via option
- Print warning if FileLogging is disabled but an argument for the logDir is present
- LogLevel is considered for all loggers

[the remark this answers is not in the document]
martin-fleck-at: Yeah, great suggestion, that sounds good!

tortmayr:
I addressed the mentioned issues and refactored the CLIParser class as well.

PR #72, opened 2020-08-16
#101 Ensure proper handling of empty DeleteElementOperation

tortmayr:
Log a warning instead of throwing an exception if no elementIds are specified for an DeleteElementAction.
Throwing an exception here will break the command stack. In addition, it's not necessary because technically an empty DeleteElementOperation is still valid (and simply does not affect the model at all)

Fixes eclipse-glsp/glsp/issues/101


martin-fleck-at (approved):
Wonderful technicalities but I agree that an exception is a bit harsh, so logging a warning is more than enough, thank you for finding and fixing this!

PR #73, opened 2020-08-24
 #96 #94 Cleanup and refactor client-server communication

tortmayr:
- Refactor client-server communication into a base protocol and a jsonrpc specific implementation (#94)
- Cleanup protocol and implement (previously unused) shutdown method
- Add ClientSessionManager to track lifecycle of GLSP client connections (#96)
- Add listener-mechanism to react to lifecycle changes (used in 
DefaultModelStateProvider)
- Remove (now obsolete) GLSPClientProxyProvider
- Fix wrong version of guava in org.eclipse.glsp.graph pom.xml
- Add unique applicationId to InitializeParameters

Part of:
- eclipse-glsp/glsp/issues/94
- eclipse-glsp/glsp/issues/96


tortmayr:
@CamilleLetavernier Could you have a look at the ?ClientSessionMananger`? Does it provide everything you need in terms of lifecycle managment?

plugins/org.eclipse.glsp.api/src/org/eclipse/glsp/api/protocol/ClientSessionManager.java
```diff
@@ -0,0 +1,35 @@
...
+
+   boolean disposeClientSession(String clientId);
+
+   boolean disconnectClient(GLSPClient client);
+
+   boolean addListener(ClientSessionListener listener);
+
+   boolean removeListener(ClientSessionListener listener);
+
+   Optional<GLSPClient> resolve(String clientId);
```
CamilleLetavernier, line 34: Is this reliable? If I understand correctly, 1 GLSPClient corresponds to one "tab" in a browser, and 1 client ID corresponds to one "Widget" in a tab. So if we have 2 different GLSPClients, they may use the same clientId (Typically, xxx-diagram_0_widget). Since (AFAIU) the ClientSessionManager is shared across multiple modules, we can't assume that clientId will be unique ids (They are only unique for 1 GLSPClient / injection Module)

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/jsonrpc/DefaultClientSessionManager.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.eclipse.glsp.api.protocol.ClientSessionListener;
+import org.eclipse.glsp.api.protocol.ClientSessionManager;
+import org.eclipse.glsp.api.protocol.GLSPClient;
+
+public final class DefaultClientSessionManager implements ClientSessionManager {
+
+   private final Set<GLSPClient> clients = new HashSet<>();
+   private final Set<ClientSessionListener> listeners = new LinkedHashSet<>();
+   private final Map<String, GLSPClient> clientSessions = new HashMap<>();
```
CamilleLetavernier, line 35: I think this doesn't work with the way clientIds are currently generated

CamilleLetavernier (commented):
The ClientSessionMananger seems to provide everything we need; however I have some concerns about the clientId/GLSPClient mapping (See comments in line)

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/jsonrpc/DefaultClientSessionManager.java
tortmayr: Works because I adapted the way that clientIds are generated. They are now unique for each actual client/application

plugins/org.eclipse.glsp.api/src/org/eclipse/glsp/api/protocol/ClientSessionManager.java
CamilleLetavernier: Ah, great! Then it looks good to me :)

CamilleLetavernier (approved):
I didn't test the changes; but the ClientSessionManager API looks good to me. I'll have to test that in real application to see if notifications are properly sent in all cases (Especially, I didn't see how/if the server can trigger disconnectClient when the connection is lost. I'm not sure that's even desired; e.g. if the client can reconnect using the same app/client Id without reloading the tab)

planger (approved):
Great work, thanks a lot!
I've appended a few minor changes in a separate commit for your review. Other than that, this looks great!
Just a general remark regarding copyrights, please keep the original where applicable but just update the year (e.g. original 2019 --> now 2019 - 2020).

plugins/org.eclipse.glsp.api/src/org/eclipse/glsp/api/action/kind/DisposeClientAction.java
```diff
@@ -1,5 +1,5 @@
-/*******************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+/********************************************************************************
+ * Copyright (c) 2020 EclipseSource and others.
```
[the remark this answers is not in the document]
tortmayr: In general you are right. But you picked a bad example. The DisposeClientAction was introduced with this commit so the copyright year should be 2020. This is displayed weirdly because Github is diffing GLSPClientAware <-> DisposeClientAction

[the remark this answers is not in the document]
tortmayr: I did not touch any existing copyright headers in this PR

PR #74, opened 2020-08-26
#97 ActionDispatcher should process all messages in the same thread

CamilleLetavernier:
- Refactor the ActionDispatcher to use an actions queue. Update the API to return CompletableFutures, as the method is now asynchronous.
- Expose the GLSPClient through the injector (Delegating to the GLSPServer)
- Fix some concurrent modification exceptions in DefaultClientSessionManager

fixes eclipse-glsp/glsp/issues/97

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/action/DefaultActionDispatcher.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/action/DefaultActionDispatcher.java
@@ -56,27 +56,27 @@ public class DefaultActionDispatcher implements ActionDispatcher, ClientSessionL
    private static final AtomicInteger COUNT = new AtomicInteger(0);
 
    @Inject
-   private ActionHandlerRegistry actionHandlerRegistry;
+   protected ActionHandlerRegistry actionHandlerRegistry;
 
    @Inject
-   private ModelStateProvider modelStateProvider;
+   protected ModelStateProvider modelStateProvider;
 
    private final ClientSessionManager clientSessionManager;
 
-   private final String name;
+   protected final String name;
 
-   private final Thread thread;
+   protected final Thread thread;
 
-   private final BlockingQueue<ActionMessage> actionsQueue = new ArrayBlockingQueue<>(100, true);
+   protected final BlockingQueue<ActionMessage> actionsQueue = new ArrayBlockingQueue<>(100, true);
 
    // Results will be placed in the map when the action dispatcher receives a new message (From arbitrary threads),
    // and will be removed from the dispatcher's thread.
-   private final Map<ActionMessage, CompletableFuture<Void>> results = Collections.synchronizedMap(new HashMap<>());
+   protected final Map<ActionMessage, CompletableFuture<Void>> results = Collections.synchronizedMap(new HashMap<>());
 
    // Use a provider, as the GLSPClient is probably not created yet. We won't receive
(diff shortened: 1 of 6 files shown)


CamilleLetavernier:
Note: since the dispatch methods are now asynchronous, existing clients using try-catch around the dispatch methods need to replace with:
```java
actionDispatcher.dispatch(...).exceptionally(ex => {
   // Handle exception
   return null;
});
```

See the changes in DefaultGLSPServer for an example

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/jsonrpc/DefaultGLSPServer.java
```diff
@@ -99,24 +99,37 @@ public void connect(final GLSPJsonrpcClient clientProxy) {
    public void process(final ActionMessage message) {
       log.debug("process " + message);
       String clientId = message.getClientId();
+
+      Function<Throwable, Void> errorHandler = ex -> {
```
tortmayr, line 103: Please extract into dedicated method that provides the errorHandler. This way subclasses can easily use a customized errorHandler

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/action/DefaultActionDispatcher.java
```diff
@@ -15,44 +15,198 @@
...
+      return dispatch(new ActionMessage(clientId, action));
+   }
+
+   @Override
+   public CompletableFuture<Void> dispatch(final ActionMessage message) {
+      if (message == null) {
+         String errorMsg = String.format("Received a null message in DefaultActionDispatcher: %s", name);
+         throw new IllegalArgumentException(errorMsg);
+      }
+      final CompletableFuture<Void> result = results.putIfAbsent(message, new CompletableFuture<Void>());
```
tortmayr, line 103: This causes  NullPointerExceptions. Result is null,if there was no previous mapping for the given key and is returned instead of valid CompletableFuture<void>. Maybe use computeIfAbsent instead

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/action/DefaultActionDispatcher.java
```diff
@@ -15,44 +15,198 @@
...
+ * </p>
+ */
+public class DefaultActionDispatcher implements ActionDispatcher, ClientSessionListener {
 
+   private static final Logger LOG = Logger.getLogger(DefaultActionDispatcher.class);
+
+   private static final AtomicInteger COUNT = new AtomicInteger(0);
+
+   @Inject
+   private ActionHandlerRegistry actionHandlerRegistry;
```
tortmayr, line 59: Please use protected instead of private for properties and methods that will potentially be used/customized by subclasses.

tortmayr (changes requested):
Thanks! Change looks good to me in general and a thread-safe Actiondispatcher is a great addition to GLSP.

I  raised two minor issues as inline comments. Other than that, everything is fine.
Just a general remark: If possible, please use protected instead of private for properties and methods. This increases extensibility and re-usability of the base framework components.

CamilleLetavernier:
Thanks for the review, Tobias! I've pushed one more commit

tortmayr (approved):
Thanks! LGTM :+1:

PR #75, opened 2020-08-27
#96 Introduce InitializeClientSessionAction

tortmayr:
Also: 
- Use new clientProxy provider instead of deriving the proxy from the client id.
- Fix/Surpress a couple of warnings
- Sort members of MultiBindingDefaults

Follow up for eclipse-glsp/glsp/issues/96

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actionhandler/DisposeClientSessionActionHandler.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actionhandler/DisposeClientSessionActionHandler.java
@@ -1,0 +1,38 @@
+/********************************************************************************
+ * Copyright (c) 2020 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0.
+ *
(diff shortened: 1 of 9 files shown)


planger (approved):
LGTM, thanks!

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actionhandler/DisposeClientSessionActionHandler.java
```diff
@@ -18,19 +18,19 @@
...
-public class DisposeClientActionHandler extends BasicActionHandler<DisposeClientAction> {
+public class DisposeClientSessionActionHandler extends BasicActionHandler<DisposeClientSessionAction> {
 
    @Inject()
    protected ClientSessionManager sessionManager;
 
    @Override
-   protected List<Action> executeAction(final DisposeClientAction action, final GraphicalModelState modelState) {
+   protected List<Action> executeAction(final DisposeClientSessionAction action, final GraphicalModelState modelState) {
       sessionManager.disposeClientSession(modelState.getClientId());
```
CamilleLetavernier, line 34: The DisposeClientSessionAction contains the ID of the session to dispose; however it's not used here (Since the model state already contains this ID, the value should be the same; but then maybe we don't need the clientId in the DisposeClientSessionAction?)

CamilleLetavernier (approved):

PR #76, opened 2020-08-28
#109 Remove the UUID requirement / ApplicationID

CamilleLetavernier:
- Remove the clientId to GLSPClient mapping in ClientSessionManager

fixes eclipse-glsp/glsp/issues/109

Signed-off-by: Camille Letavernier <cletavernier@eclipsesource.com>

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/jsonrpc/DefaultClientSessionManager.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/jsonrpc/DefaultClientSessionManager.java
@@ -48,7 +48,7 @@ public final class DefaultClientSessionManager implements ClientSessionManager {
    @Override
    public synchronized boolean createClientSession(final GLSPClient client, final String clientId) {
       connectClient(client);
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/jsonrpc/DefaultClientSessionManager.java
```diff
@@ -51,17 +48,17 @@ public synchronized boolean connectClient(final GLSPClient client) {
    @Override
    public synchronized boolean createClientSession(final GLSPClient client, final String clientId) {
       connectClient(client);
-      boolean success = clientSessions.putIfAbsent(clientId, client) == null;
+      boolean success = clientSessions.computeIfAbsent(client, c -> new HashSet<>()).add(clientId);
```
tortmayr, line 51: If you use computeIfAbsent here, we can omit the connectClient(client) call in line 50

CamilleLetavernier: We still want to notify the session listeners; but we can change the computeIfAbsent with a get()

tortmayr (approved):

PR #79, opened 2020-09-02
#111 Restructure workflow server example

tortmayr:
-Add workflow example directly to this repository as dev example
-Add Example category to p2 repository
-Cleanup deprecated references to EclipseSource
-Also set compliance level to Java11

Part of eclipse-glsp/glsp/issues/111


martin-fleck-at (approved):
The changes work very well for me and I could start both servers and open all examples as expected. Since there will be a follow-up of this task anyway and the risk of merge conflicts is very high, I believe you could merge it as is. 

Still, I just noticed a few things which should be considered, none of them are blocking though:
- The '.classpath' file from the 'org.eclipse.glsp.example.workflow' project should have the 'src-gen' ordered below the JRE as it leads to a weird display in Eclipse. Furthermore, Eclipse added a `<attribute name="module" value="true"/>` to the attributes, not sure if this is necessary but I wanted to mention it.
- The build.properties from the 'org.eclipse.glsp.example.workflow' project should have 'src' added to the sources and 'bin' removed from the output.
- VS Code workspace still needs to be updated correctly.
- Opening the workflow example in the client did not work using the task, there seems to be an 's' from 'examples' missing in the path.
- Unlinking glsp-client packages did not clean all links maybe? In VS Code they show up a s soft links but in the .config/yarn/link directory all links are gone. Not sure what is up with that.
- When building the m2 and p2 repositories I got some warnings. Specifically for the m2 build the 'src-gen' folder was not ignored by checkstyle which lead to a lot of warnings. Other than that there were some warnings which are not caused by your changes but maybe can be fixed if you are working on that anyway since the fixes are very simple:
```
[WARNING] /home/martin/Git/OpenSource/eclipse-glsp/glsp-server/plugins/org.eclipse.glsp.layout/src/org/eclipse/glsp/layout/ElkLayoutEngine.java:[76] 
	* {@link org.eclipse.elk.alg.layered.options.LayeredMetaDataProvider}.
	         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Javadoc: org.eclipse.elk.alg cannot be resolved to a type
```
--> Fix Link in JavaDoc or remove link all together

```
[WARNING] /home/martin/Git/OpenSource/eclipse-glsp/glsp-server/plugins/org.eclipse.glsp.layout/src/org/eclipse/glsp/layout/GLSPLayoutConfigurator.java:[98] 
	} else {
         return super.overrideWith(other);
      }
	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Statement unnecessarily nested within else clause. The corresponding then clause does not complete normally
```
--> Unnecessary else clause, just remove it.

```
[WARNING] /home/martin/Git/OpenSource/eclipse-glsp/glsp-server/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/WorkflowGLSPModule.java:[68] 
	protected Class<? extends GLSPServer> bindGLSPServer() {
	                          ^^^^^^^^^^
GLSPServer is a raw type. References to generic type GLSPServer<T> should be parameterized
```
--> Simply suppress the warning.

In any case, great work and thank you very much Tobias!

PR #80, opened 2020-09-24
#123 Centralize configuration in the server

CamilleLetavernier:
When receiving an InitializeClientSessionAction, the server now sends a ConfigureServerHandlersAction which contains the list of actionKinds that the server can handle.

The client can use this to dynamically configure server-handled actions, rather than listing them explicitly.


planger (approved):
Thank you very much! An important steps towards letting the custom server configure a more and more generic client.

PR #81, opened 2020-09-30
Refactor internal and public API

planger:
Resolves https://github.com/eclipse-glsp/glsp/issues/1


CamilleLetavernier:
The M2 build fails because `org.eclipse.glsp.layout/pom.xml` still references `org.eclipse.glsp.api`

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/util/ReflectionUtil.java
```diff
@@ -13,7 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-package org.eclipse.glsp.api.utils;
+package org.eclipse.glsp.server.internal.util;
```
CamilleLetavernier, line 16: Since we encourage reflection in the GLSPModule (By only declaring the Handler, and making the handled Operation implicit), I think client applications might need this helper as well. I'm not 100% sure though; as the use case we had in Ecore was a temporary workaround, that I can probably revert now (We had to provide a custom version of the DIOperationHandlerRegistry)

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/util/LayoutUtil.java
```diff
@@ -13,17 +13,13 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ******************************************************************************/
-package org.eclipse.glsp.api.utils;
+package org.eclipse.glsp.server.internal.util;
```
CamilleLetavernier, line 16: Should Util classes really be internal? Especially this one that can be used to manipulate the model, which is probably what client applications should focus on

CamilleLetavernier:
I've started migrating Ecore GLSP to these changes and it was almost painless. The main changes I noticed:

- org.eclipse.glsp.api merged with org.eclipse.glsp.server (Obviously, that was the point of this task :) )
- Package Names changed accordingly; can usually be auto-solved by using Eclipse's organize imports
- GraphicalModelState renamed to GModelState
- ServerConfiguration changed to ServerLayoutConfiguration and DefaultGLSPModule#bindServerConfiguration renamed to DefaultGLSPModule#bindServerLayoutConfiguration
- MultiBindConfig renamed to MultiBinding

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/Registry.java
```diff
@@ -13,7 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-package org.eclipse.glsp.api.registry;
```
CamilleLetavernier, line 16: This interface is (indirectly) exposed by all the public Registries. While it's not necessary to make it non-internal (As it's not supposed to be used directly), I wonder if making it internal is really a good idea. Isn't there a risk of changing this interface (which should be safe, since it's internal), and then breaking the API for all non-internal registries? (ActionRegistry, etc). 

Again, there's only one such occurrence in Ecore GLSP, and it's related to the no-longer necessary workaround that provides a custom (Ecore)DIOperationHandlerRegistry, so maybe we don't expect registries to be customized anyway. But it feels weird to have public/exposed methods declared in an internal interface.

planger: You're right, it doesn't really make sense since they are sort-of public API anyway.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/util/LayoutUtil.java
planger: Hm, not sure.. I think those utilities that are not made for being used in the main use cases should be internal to minimize what we have to be careful to change. But I agree, if one _might_ be useful, it is the LayoutUtil.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/util/ReflectionUtil.java
planger: I'd say that this is util that typical adopters shouldn't have to use. I agree that it might be useful in certain cases, but at the same time I'd like to minimize the public API to avoid overhead when changing the API in the future.
What do you think, @martin-fleck-at and @tortmayr?

tortmayr: I agree, this s a util that typical adopters shouldn't  have to use. However, I don't have a strong opinion wether we should make it public or not. I can see that it might be useful in some edge cases. And I don't think that the `ReflectionUtil` will see much changes in the future so the overhead should be minimal as well. I'm fine with both keeping it as is or exposing it.

martin-fleck-at: I feel the same as Tobias although I am slightly tilting towards making it public as the API will be very stable and it might be useful for some adopters to use it instead of writing the same code again. Of course I am aware that this is not "core" GLSP and making this "our" API is a bit unexpected. So I am fine with anything in this regard, I am sure that helped a lot :P

CamilleLetavernier: In this case, the middle-ground is typically to export the package, but still mark it internal / hidden, so clients can use it with a nice warning. In this case, we don't have to guarantee the API, and clients that really need it have to think carefully about what they're doing; but still they don't have to rewrite it if they use it. And if it changes in the future, these clients would likely break anyway (Whether they use the exact same class, or a copy of it), as it probably means we changed the way e.g. Handlers are injected.

planger: Yes, I'd actually do that for all non-API classes; that is exporting them, but marking them as internal. There is no reason to really prevent an adopter to use those classes, but it is merely to communicate that this class may change anytime.
Thanks all for your feedback!

martin-fleck-at (approved):
Let's merge this for now and resolve any potential issues afterwards as such a change has a huge conflict potential ;-) Thank you very much for your hard work, Philip!

tortmayr:
Great change! I think the capability-based package structure is a big improvement. I only noticed a couple of minor  "issues". Maybe we should briefly discuss them here and then great a follow-up ticket for this.

- The `org.eclipse.glsp.server` package doesn't follow the capability-based approach it rather serves as a container for miscellaneous that didn't fit anywhere else. Maybe we should move `GLSPModule` and `DefaultGLSPModule` into the "di" subpackage and the other classes into the "util" subpackage?

- `org.eclipse.glsp.server.factory` package simply groups classes that have the "Factory" affix but are not necessary related to each other. I'd suggest to move the `PopupmodelFactory` into  `org.eclipse.glsp.server.features.popup` and create a new feature package for model loading where we group the RequestModelAction, the corresponding handler and the ModelFactories

- Maybe we should also create a feature subpackage for undo/redo.

- I think it would be a good idea to move all actions that are declared as `ClientAction`(https://github.com/eclipse-glsp/glsp-server/blob/master/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/di/MultiBindingDefaults.java#L101)  into a dedicated subpackage.

PR #83, opened 2020-10-27
Add model source change notification

planger:
A model source change notification is a message sent from the server to
the client to indicate that the original source of the model has
changed. Clients may choose to reload the diagram from the changed
source (RequestModelAction) or ignore this notification and overwrite
the external changes on the next save (SaveAction).

https://github.com/eclipse-glsp/glsp/issues/134


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SaveModelActionHandler.java
```diff
@@ -56,6 +61,7 @@ private void saveModelState(final GModelState modelState) {
             LOG.error(e);
          }
       });
+      modelSourceWatcher.continueWatching(modelState);
```
martin-fleck-at, line 64: Should we do that in a try-finally-block or do we assume that if there is an error when saving the model state we also want to stop watching the model source?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/FileWatcher.java
```diff
@@ -0,0 +1,195 @@
...
+
+   @Override
+   public void continueWatching(final GModelState modelState) {
+      getAllWorkers(modelState.getClientId()).forEach(FileWatchWorker::continueNotifications);
+   }
+
+   protected List<Path> getPaths(final GModelState modelState) {
+      final Optional<String> uriString = ClientOptions.getValue(modelState.getClientOptions(),
+         ClientOptions.SOURCE_URI);
+      URI sourceUri = URI.createFileURI(uriString.map(uri -> uri.replace(FILE, "")).orElseThrow());
```
martin-fleck-at, line 96: I am not a big fan of this URI manipulation hidden in this method. There is something similar going on in the JsonFileModelFactory (convertToFile). Do you think it would make sense to extract that into a utility class somewhere or even put it into the ClientOptions as a special method since they already define the constants for that anyway. Same goes for the FILE constant.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/ModelSourceWatcher.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+package org.eclipse.glsp.server.features.modelsourcewatcher;
+
+import org.eclipse.glsp.server.model.GModelState;
+
+public interface ModelSourceWatcher {
+
+   default void startWatching(final GModelState modelState) {}
```
martin-fleck-at, line 22: Please add some documentation above those methods. To me it is not obvious what each of those methods does and how they interact, e.g., what is the difference between starting and stopping or pausing and continuing? Is there any expected behavior when you call 'continueWatching' before it has been started, etc.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/FileWatcher.java
```diff
@@ -0,0 +1,195 @@
...
+   }
+
+   @Override
+   public void sessionClosed(final String clientId, final GLSPClient client) {
+      disposeAllWorkers(clientId);
+   }
+
+   @Override
+   public void startWatching(final GModelState modelState) {
+      createWorkers(modelState).forEach(FileWatchWorker::start);
```
martin-fleck-at, line 75: Not sure how expensive that is but would it make sense to check whether the existing workers are already watching the respective files?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/RequestModelActionHandler.java
```diff
@@ -29,18 +30,23 @@
...
+   private ModelSourceWatcher modelSourceWatcher;
+
    @Override
    public List<Action> executeAction(final RequestModelAction action, final GModelState modelState) {
-
       GModelRoot model = modelFactory.loadModel(action, modelState);
       modelState.setRoot(model);
       modelState.setClientOptions(action.getOptions());
+
+      modelSourceWatcher.startWatching(modelState);
```
martin-fleck-at, line 42: It seems a bit asymmetrical that we start the watching here but we never actually stop the watching but do so by having the watcher listen to the client session. I know it is like this because the client options are only sent through the request model action so this is the earliest we can actually start the watching. We could extend the ClientSessionManager with an 'updateClientSession' or something similar so that the whole file watcher could be an encapsulated class. But this is just a random thought and does not have to be done in this PR ;-)

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/FileWatcher.java
```diff
@@ -0,0 +1,195 @@
...
+      return Optional.ofNullable(workers.get(clientId)).stream().flatMap(Collection::stream);
+   }
+
+   private List<FileWatchWorker> createFileWatchWorkers(final GModelState modelState) {
+      return getPaths(modelState).stream().map(path -> new FileWatchWorker(modelState.getClientId(), path))
+         .collect(Collectors.toList());
+   }
+
+   protected void notifyClient(final String clientId, final Path filePath) {
+      actionDispatcher.dispatch(clientId, new ModelSourceChangedAction(filePath.toAbsolutePath().toString()));
```
martin-fleck-at, line 129: I noticed that sometimes we get two changes for the same file in quick succession. Maybe we should add a debounce here to avoid sending the same notification twice? Alternatively we can argue that the client must handle this.

martin-fleck-at (commented):
Change works for me, thank you very much Philip! Everything looks really great and I especially LOVE the additional test cases!

Regarding the code I have a few minor questions and suggestions but everything works as expected for me. Of course the client side is still missing so the integration will also be interesting ;-)

martin-fleck-at:
If the client change is also ready please do not forget to provide an additional PR that extend the GLSP protocol with the new `ModelSourceChangedAction` :wink:

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SaveModelActionHandler.java
planger: Good idea, done!

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/FileWatcher.java
planger: Yes, I agree! I added a static method to ClientOptions and harmonized those two methods.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/ModelSourceWatcher.java
planger: Agreed, done!

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/FileWatcher.java
planger: Imho this would complicate the thread management as we now create all workers on start and always start from a clean slate if invoked again. I don't think it is expensive, but I haven't measured it.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/RequestModelActionHandler.java
planger: Right, I agree! Hm, but I feel this needs more thought. So I'd rather prefer doing that in a follow up PR.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/FileWatcher.java
planger: Yes, definitely! I added a debouncer.

planger:
Thanks a lot for your review! Great feedback! I've addressed most comments and replied to the ones that haven't been addressed yet. Thanks!

martin-fleck-at:
Thank you very much Philip! I did some minor fixes but everything works great now!

martin-fleck-at (approved):

PR #84, opened 2020-10-29
#140 Properly use completable future during server initialization

tortmayr:
This way ensures that the server is only considered "initialized" after the options have  been processed and handled.
Fixes eclipse-glsp/glsp/issues/140


martin-fleck-at (approved):
Looks good to me and works without any issues, thank you Tobias!

PR #86, opened 2020-11-05
#124 [Server - ActionDispatcher] Missing error handling for responses

CamilleLetavernier:
- Aggregate the CompletableFutures from ActionDispatcher#dispatchAll into a single CompletableFuture
- Report this as the result of the Action

Several cases are then possible:

- If an action doesn't have any side-effect ("responses"), the behavior is the same as before
- If an action has a single side-effect, use its result as the action's result
- If an action has multiple side-effects, wait until all of them are complete, then aggregate the exceptions (If any) to complete exceptionally (Otherwise, simply complete the future).

To test this, simply throw exceptions in e.g. ClientActionHandler.send() (As side-effects are typically reply-actions sent to the Client). Without this patch, these actions are not logged/reported. With the patch, the first exception is reported (Other exceptions are suppressed, but can still be accessed via Exception#getSuppressed() if necessary).

Note that the patch introduces a bit of support for asynchronous execution (i.e. instead of processing the results directly in #handleMessage, we now have to wait until all side effects are complete). Since (AFAIK) we don't have asynchronous action handlers in GLSP, this can't be easily tested; but this should already be supported with this patch.

Fixes eclipse-glsp/glsp/issues/124

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/action/DefaultActionDispatcher.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/action/DefaultActionDispatcher.java
@@ -15,6 +15,7 @@
  ********************************************************************************/
 package org.eclipse.glsp.server.internal.action;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


tortmayr (approved):
LGTM. :+1:  Thanks

PR #87, opened 2020-11-06
#140 Improve extendibility of DefaultGLSPServer

tortmayr:
Avoid private members in the default implementation for easier customization in  subclasses.

Follow-up for eclipse-glsp/client/issues/140


martin-fleck-at (approved):
Looks good to me, thank you Tobias!

PR #35, opened 2020-01-20
glsp#12 Add initial support for copy & paste

planger:
https://github.com/eclipse-glsp/glsp/issues/12

To test this change, you'll also need the following changes:
  * eclipse-glsp/glsp-examples#25
  * eclipse-glsp/glsp-client#26
  * eclipse-glsp/glsp-theia-integration#16

PR #88, opened 2020-11-17
#35 Extend server launcher for embedded use case

tortmayr:
Extend the `DefaultGLSPServerLauncher` to print a message to the output stream once the startup is completed and the server is ready to accept client requests.
This is needed for usecases where the client launches the server itself, in an embedded process. This way the client process can wait for the startup success message from the server before trying to connect to it.

Also: Fix checkstyle warings in 'FutureUtil'

Part of eclipse-glsp/glsp/issues/35


martin-fleck-at (approved):
Looks good to me, thank you Tobias!

PR #89, opened 2020-11-18
#132 Further clean up server API

tortmayr:
- Move  classes from root package into a matching subpackage
- Create "org.eclipse.glsp.server.features.core.model" package that contains all actions and related classes that are needed for model loading
- Create a dedicated feature subpackage for undo/redo

Resolves eclipse-glsp/glsp/issues/132


martin-fleck-at (approved):
Great work, thank you Tobi! Change looks good to me and discussed separately adding a dedicated package for the client actions (as proposed in the issue) is not useful at the moment since we are already splitting actions according to features.

PR #90, opened 2020-11-24
#122 #123 Centralize configuration in the server

CamilleLetavernier:
- Merge ServerLayoutConfiguration into DiagramConfiguration
- Centralize model update protocol in Model Submission Handler

This addresses both eclipse-glsp/glsp/issues/122 and eclipse-glsp/glsp/issues/123, as the changes to needsClientLayout affects the way the ModelSubmissionHandler is implemented.


CamilleLetavernier:
Note: I also have the corresponding client-side changes (Which simply removes the needsClientLayout parameter from the client); I'll create a separate PR if these changes are fine. The client-side changes aren't actually required (strictly speaking), as the extra parameter would simply be ignored by the server.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
```diff
@@ -34,21 +38,73 @@
...
+    *
+    * @param update
+    *                      <code>true</code> if this is an update to an existing model; <code>false</code>
+    *                      if this is a new model (e.g. after loading a model)
+    * @param modelState
+    *                      The {@link GModelState}
+    * @return
+    *         A list of Actions to be processed in order to submit the model
+    */
    public List<Action> doSubmitModel(final boolean update, final GModelState modelState) {
```
tortmayr, line 64: Maybe just call it `submitModel`?

tortmayr (changes requested):
Looks good to me. I only have to minor change requests :wink:

planger:
This looks very useful to me! Can we get this merged soon?
I'd like to address https://github.com/eclipse-glsp/glsp/issues/119 based on this change too, because the model submission handler could just invoke an extended version of the model factory to translate model state (with the custom model source) into the new version of the gmodel before submitting it to the client.
This would then also centralize the translation from original model source to gmodel.

planger:
Closed in favor of https://github.com/eclipse-glsp/glsp-server/pull/95

PR #91, opened 2020-11-26
149: [Websocket] Do not reuse the same Injector for each client connection

CamilleLetavernier:
- Create a new Injector instance for each client connection, as expected by the GLSP Server infra

fixes eclipse-glsp/glsp/issues/149


tortmayr (approved):
:+1: LGTM. Thanks

PR #94, opened 2020-12-07
#156 Use package-import rather than require-bundle for javax.servlet

CamilleLetavernier:
- Multiple bundles may provide javax.servlet packages, which can result in some conflicts at runtime. Using a strict package-import, with the same dependency range as Jetty, seems more reliable.


tortmayr (approved):
LGTM. :+1:

PR #98, opened 2021-02-15
#191: Provide correct kind for SelectAction

martin-fleck-at:
Fixes https://github.com/eclipse-glsp/glsp/issues/191


martin-fleck-at:
It is even correct in the PROTOCOL definition: https://github.com/eclipse-glsp/glsp/blob/master/PROTOCOL.md#2931-selectaction ;-)

planger (approved):

PR #100, opened 2021-02-23
Add args to SModelElements

planger:
This allows to transfer custom information from the server to the client
in the GModel without having to extend the GModel types with custom
classes.

E.g.
    modelState.getRoot().getArgs().put("test", true);
    modelState.getRoot().getArgs().put("test2", "a string 2");

https://github.com/eclipse-glsp/glsp/issues/194


tortmayr (commented):
Looks good to me. However, we should probably update the Copyright headers of the generated files (2019-2021).

tortmayr (approved):

PR #101, opened 2021-03-22
#197 #196 Improve dirty state handling

tortmayr:
The `SetDirtyStateAction` now also declares the action/operation which caused the dirty state change. 
Ensure that the `SetDirtyStateAction` is only sent once during a model update
Ensure that layout engine is invoked after bounds have been computed (AutomaticServerLayout case)

Part of https://github.com/eclipse-glsp/glsp/issues/197
Fixes https://github.com/eclipse-glsp/glsp/issues/196

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SetDirtyStateAction.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SetDirtyStateAction.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2020 EclipseSource and others.
+ * Copyright (c) 2020-2021 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -20,18 +20,24 @@ public class SetDirtyStateAction extends Action {
    public static final String ID = "setDirtyState";
 
    private boolean isDirty;
+   private Action causedBy;
 
    public SetDirtyStateAction() {
       super(ID);
    }
 
-   public SetDirtyStateAction(final boolean isDirty) {
+   public SetDirtyStateAction(final boolean isDirty, final Action causedBy) {
       this();
       this.isDirty = isDirty;
+      this.causedBy = causedBy;
    }
 
    public boolean isDirty() { return isDirty; }
 
    public void setDirty(final boolean isDirty) { this.isDirty = isDirty; }
+
+   public Action getCausedBy() { return causedBy; }
+
+   public void setCausedBy(final Action causedBy) { this.causedBy = causedBy; }
 
 }
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SetDirtyStateAction.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SetDirtyStateAction.java
@@ -16,28 +16,39 @@
 package org.eclipse.glsp.server.actions;
 
 public class SetDirtyStateAction extends Action {
-
    public static final String ID = "setDirtyState";
 
    private boolean isDirty;
-   private Action causedBy;
+   private String reason;
 
    public SetDirtyStateAction() {
       super(ID);
    }
 
-   public SetDirtyStateAction(final boolean isDirty, final Action causedBy) {
+   public SetDirtyStateAction(final boolean isDirty) {
       this();
       this.isDirty = isDirty;
-      this.causedBy = causedBy;
+   }
+
+   public SetDirtyStateAction(final boolean isDirty, final String reason) {
+      this(isDirty);
+      this.reason = reason;
    }
 
    public boolean isDirty() { return isDirty; }
 
    public void setDirty(final boolean isDirty) { this.isDirty = isDirty; }
 
-   public Action getCausedBy() { return causedBy; }
+   public String getReason() { return reason; }
 
-   public void setCausedBy(final Action causedBy) { this.causedBy = causedBy; }
+   public void setReason(final String reason) { this.reason = reason; }
+
+   public static class Reason {
+      public static final String OPERATION = "operation";
+      public static final String UNDO = "undo";
+      public static final String REDO = "redo";
+      public static final String SAVE = "save";
+      public static final String EXTERNAL = "external";
+   }
 
 }
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+ * Copyright (c) 2019-2021 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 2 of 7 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SetDirtyStateAction.java
```diff
@@ -20,18 +20,24 @@
...
 
    private boolean isDirty;
+   private Action causedBy;
 
    public SetDirtyStateAction() {
       super(ID);
    }
 
-   public SetDirtyStateAction(final boolean isDirty) {
+   public SetDirtyStateAction(final boolean isDirty, final Action causedBy) {
```
planger, line 29: Do we have to do this a mandatory parameter? Afaik, VSCode just wants to know that for undo actions, right? I'm wondering whether this justifies making this a mandatory parameter for all other invocations caused by anything else than `UndoAction`. Also do you think it would be simpler to just have a string argument: `reason = 'operation' | 'undo' | 'save'`? I don't have a hard opinion on that, but enforcing to provide an actual action seems like a strong requirement for this rather specific use case of VSCode.

Minor knitpick, I think as a name I'd prefer `reason` rather than `causedBy`.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
```diff
@@ -87,15 +89,44 @@
...
     */
-   public List<Action> submitModelDirectly(final GModelState modelState) {
+   public List<Action> submitModelDirectly(final GModelState modelState, final Action causedBy) {
       GModelRoot gModel = modelState.getRoot();
-      Action action = gModel.getRevision() == 0 ? new SetModelAction(gModel) : new UpdateModelAction(gModel);
+      DiagramConfiguration diagramConfiguration = diagramConfigurationRegistry.get(modelState);
+      if (diagramConfiguration.getLayoutKind() == ServerLayoutKind.AUTOMATIC) {
+         layoutEngine.layout(modelState);
+      }
+      Action modelAction = gModel.getRevision() == 0 ? new SetModelAction(gModel) : new UpdateModelAction(gModel);
```
martin-fleck-at, line 101: I already had cases where I had to override this method only to change the `animate` flag in this method. Would it be possible to extract the `new UpdateModelAction(gModel)` into a separate method or provide a method that returns the animate boolean that is then used in the creation of this action? I know it is not strictly part of this change but with this change the method just became more complicated and overriding more cumbersome ;-)

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
```diff
@@ -87,15 +89,44 @@
...
       synchronized (modelLock) {
-         return Arrays.asList(action, new SetDirtyStateAction(modelState.isDirty()));
+         List<Action> result = new ArrayList<>();
+         result.add(modelAction);
+         Optional.ofNullable(causedBy).ifPresent(result::add);
```
martin-fleck-at, line 105: I'm not sure I understand why we would need to add the 'cause' to the result? The method comment mentions that the result of this call should be dispatched through the ActionDispatcher (either directly or not), would that lead not to an infinite loop in case we do not need a client layout or do I miss something here?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
```diff
@@ -55,28 +57,28 @@
...
+      boolean needsClientLayout = diagramConfiguration.needsClientLayout();
+      if (needsClientLayout) {
          synchronized (modelLock) {
             return Arrays.asList(new RequestBoundsAction(modelState.getRoot()),
-               new SetDirtyStateAction(modelState.isDirty()));
+               new SetDirtyStateAction(modelState.isDirty(), causedBy));
          }
       }
-      return submitModelDirectly(modelState);
+      return submitModelDirectly(modelState, needsClientLayout ? null : causedBy);
```
martin-fleck-at, line 74: It seems like it should not be possible for `needsClientLayout` to be true here. If it were, we would already have returned.

martin-fleck-at (changes requested):
Hi Tobi, in our example everything works fine and well but I am not sure this is true if we do not need a client layout. Maybe I am missing something and you can clarify that case for me (see comments inline).

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SetDirtyStateAction.java
martin-fleck-at: +1 for `reason` ;-)

tortmayr: +1 for 'reason' and making it optional. In any case it should be possible to determine wehter the dirty state change was triggered by an actual new edtit/operation, an undo/redo action or a save action. Now thinking about it, we also have a another case: the server could also update the model without an operation invocation (e.g. by receiving an update from the model server).
So maybe we should type reason as follows: 'operation'|'undo'|'redo'|'save'|'externalUpdate'.
What do you think?

planger: :+1: Yes, sounds good! Maybe just `external` is good enough, e.g. also if the file got updated on the file system or whatever other external reason. Thanks!

martin-fleck-at: Maybe it should be just String then and adopters can provide any reason they want and we just have a list of known String reasons but no strict typing?

tortmayr: Fine with me as well. We could then just use the action.kind as default reason string as well

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
tortmayr: I'm wondering if this is something that we could store in the `DiagramConfiguration` (similar to the needsClientLayout flag).  This way the user doesn't have to subclass the `ModelSubmissionHandler` and can configure it directly in the DiagramConfiguration.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
```diff
@@ -55,28 +56,32 @@
     * </p>
     *
     * @param modelState The model state to submit.
+    * @param reason     The action/operation that caused the model update
```
martin-fleck-at, line 59:
```suggestion
    * @param reason     The optional reason that caused the model update.
```
I believe this is still out-dated.

martin-fleck-at (changes requested):
Changes look good to me and as far as I can tell, everything still works as expected, thank you very much Tobias! I just have one minor request to update an outdated Javadoc.

martin-fleck-at (approved):
Thank you, everything looks good to me now!

PR #103, opened 2021-03-24
#208 Add `fileUri` to `SaveModelAction`

tortmayr:
Also update `SaveModelActionHandler` to process the new property accordingly.
If a fileUri has been passed that is not equal to the sourceURi, the handler will save the current modelstate to the new fileUri but keeps the current dirty state
Part of eclipse-glsp/glsp/issues/208


martin-fleck-at (approved):
Change looks good to me and everything still works as expected. Tested with a fileUri set and it behaved as it should, thank you Tobias!

PR #108, opened 2021-05-06
#233 Add operation arguments to CreateNodeOperationHandler

tortmayr:
Resolves eclipse-glsp/glsp/issues/233

### DIFF
--- a/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/handler/CreateActivityNodeHandler.java
+++ b/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/handler/CreateActivityNodeHandler.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+ * Copyright (c) 201-2021 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -15,6 +15,7 @@
  ********************************************************************************/
 package org.eclipse.glsp.example.workflow.handler;
 
+import java.util.Map;
 import java.util.Optional;
 
 import org.eclipse.glsp.example.workflow.utils.ModelTypes;
@@ -43,7 +44,8 @@ public abstract class CreateActivityNodeHandler extends CreateWorkflowNodeOperat
    }
 
    @Override
-   protected GNode createNode(final Optional<GPoint> point, final GModelState modelState) {
+   protected GNode createNode(final Optional<GPoint> point, final Map<String, String> args,
+      final GModelState modelState) {
       return builder(point, modelState).build();
    }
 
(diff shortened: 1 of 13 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/operations/gmodel/CreateNodeOperationHandler.java
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+ * Copyright (c) 2019-2021 EclipseSource and others.
```
martin-fleck-at, line 2: The year of the other changed files was not updated. I'd suggest to either to it for all or none of them for consistency.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/operations/gmodel/CreateNodeOperationHandler.java
```diff
@@ -40,13 +41,13 @@ public void executeOperation(final CreateNodeOperation operation, final GModelSt
          container = Optional.of(modelState.getRoot());
       }
 
-      GModelElement element = createNode(getLocation(operation), modelState);
+      GModelElement element = createNode(getLocation(operation), operation.getArgs(), modelState);
```
martin-fleck-at, line 44: As far as I can see the map may be null here. I am not too much of a fan of null collection parameters, especially when point is wrapped in a nice Optional. But it could be that this is consistent with other operation handlers as well, just wanted to mention that in case it is different anywhere else.

martin-fleck-at (changes requested):

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/operations/gmodel/CreateNodeOperationHandler.java
tortmayr: Good point. I updated the `CreateNodeOperation` so that the args map is initalized with an empty map. This way we can avoid the overhead of wrapping the Map into an `Optional` and the map should always be defined. Unless somebody explicitly uses the setter to set the map to null, which should hopefully never be the case :wink:)

martin-fleck-at: I like having an empty map before having an optional map much more! However, it seems that CreateEdgeOperation and CreateNodeOperation both initialize the `args` with null due to re-using the full constructor. So either we do a `putAll` in the `CreateNodeOperation` constructor for non-null maps OR the subclasses also need to use an empty map as their default constructor argument.

tortmayr: Good catch. I checked other occurrences where we use an arbitrary args Map (e.g. EditorContext, RequestModelAction etc.) and we always more or less operate on the assumption that this map is not null. So I updated the constructors of all classes that use an args Map to ensure that they are initialized with an empty map.

examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/handler/CreateActivityNodeHandler.java
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+ * Copyright (c) 201-2021 EclipseSource and others.
```
martin-fleck-at, line 2:
```suggestion
 * Copyright (c) 2019-2021 EclipseSource and others.
```

martin-fleck-at (changes requested):

martin-fleck-at (approved):

PR #117, opened 2021-07-15
Replace `mvn install` with `mvn verify`

planger:
This avoids the installing the build artifacts into the local repository which is actually unnecessary and may also lead to problems.

https://github.com/eclipse-glsp/glsp/issues/259


tortmayr (approved):
:+1:

PR #119, opened 2021-08-03
#260 Update to Guice 5.0.1

tortmayr:
Update the dependencies to use the latest Guice version available in Orbit (5.0.1) and update the target platform definition to Eclipse 2021-03 to be in line with the TPD of the eclipse-integration.

This indirectly also enforces updates of jupiter, elk and guava dependencies.
This PR contains a workaround to bypass the guava version restriction for elk maven dependencies that can be removed once ELK 0.8.0 is released (this is tracked in  eclipse-glsp/glsp/issues/#265)

Resolves eclipse-glsp/glsp/issues/260


planger (approved):

PR #123, opened 2021-08-12
#315 Refactor base protocol

tortmayr:
Refactor the base communication protocol to support initializing and disposing a client session for a given language. The method parameters are encapsulated in dedicated parameter objects to provide a stable API. Refactor the existing `initialize` method to return void instead of a boolean and use a generic args map for custom arguments instead of a dedicated object. In addition, the protocol version that is implemented by the client is now passed as part of the `InitializeParameters`

Other fixes/improvements:
- Fixed code template for new files. Used  ${year} variable for the copyright header instead of hard coded 2020 string.
- Refactor `ActionRegistry` to use Class<Action> instead of dedicated action instances.
- Fixed bug in "Disposable". Previously the 'disposed' flag was not set correctly
- Refactor `ClientOptions` util class and extract a generic `MapUtil`  class.

Part of eclipse-glsp/glsp/issues/315


planger (approved):

planger:
Thanks for the great work! Looks good, local build and tests work fine!

tortmayr:
Thanks for the review Philip, I have made some additional changes to the protocol. Could you please re-review?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/Action.java
```diff
@@ -15,8 +15,20 @@
  ********************************************************************************/
 package org.eclipse.glsp.server.actions;
 
+/**
+ * Java-implementation of the sprotty `Action` interface. An action is a declarative description of a behavior that
```
planger, line 19:
```suggestion
 * Java-implementation of the `Action` interface. An action is a declarative description of a behavior that
```

I have the feeling to a user it is confusing to consider Action as a Sprotty concept, since it is used as a base mechansim for the entire GLS protocol too.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/ActionMessage.java
```diff
@@ -15,9 +15,17 @@
  ********************************************************************************/
 package org.eclipse.glsp.server.actions;
 
+/**
+ * Java-implementation of the sprotty `ActionMessage` interface.
```
planger, line 19:
```suggestion
 * Java-implementation of the `ActionMessage` interface.
```

planger (approved):

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/Action.java
tortmayr: I agree, however, I aligned this with the protocol.md file in eclipse-glsp/glsp. So we should probably change it there as well

PR #128, opened 2021-09-08
#385 Improve customizability of DefaultGLSPServerLauncher

tortmayr:
- Change visibility of `createClientConnection()` to protected
- Extract creation of gsonConfigurator and messageWrapper into submethods so that  they can be easily overwritten.

Fixes eclipse-glsp/glsp/issues/385


martin-fleck-at (approved):
Always good to have better customizability, thank you Tobias!

PR #135, opened 2021-10-19
#425#120 Rework Actionhandler/Operationhandler API

tortmayr:
With the DI rework #127 its no longer necessary to pass the model state object as part of the execute method. instead action/operation handlers
can simply directly inject the model state if the want to use it.  Since action & operation handlers are now client session scoped an arbitary state object can be injected and used for execution. (Part of eclipse-glsp/glsp/issues/120)

- Remove the modelstate argument from the execute() method of `ActionHandler` and `OperationHandler`
- Remove `Handler` super interface because it doesn't provide any additional value
- Add documentation for affected API.
- Introduce new `DefaultActionHandler` & `DefaultOperationHandler` that serve as replacement for the now deprecated `BasicActionHandler` & `BasicOperationHandler`
- Deprecate `BasicOperationHandler` & `BasicActionHandler` & `BasicCreateOperationHandler`
- Replace usage of `BasicActionHandler` with `DefaultActionHandler`
- Replace usage of `BasicOperationHandler'  with `DefaultActionHandler`
- Replace usage of `BasicCreateOperationhandler'  with `DefaultOperationHandler` & `DefaultCreateOperationHandler`

Fixes eclipse-glsp/glsp/issues/425

### DIFF
--- a/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/WorkflowPopupFactory.java
+++ b/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/WorkflowPopupFactory.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+ * Copyright (c) 2019-2021 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -28,7 +28,6 @@ import org.eclipse.glsp.graph.GPreRenderedElement;
 import org.eclipse.glsp.graph.GraphFactory;
 import org.eclipse.glsp.server.features.popup.PopupModelFactory;
 import org.eclipse.glsp.server.features.popup.RequestPopupModelAction;
-import org.eclipse.glsp.server.model.GModelState;
 
 public class WorkflowPopupFactory implements PopupModelFactory {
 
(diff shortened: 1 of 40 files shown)


tortmayr:
> I haven't really made up my mind yet, but we could even go as far as injecting also the action / operation with the concrete type. This could make several abstract classes superfluous. But it would also make the interface of handlers a bit implicit. WDYT?
I'm a bit lost here. I don't how injecting the action/operation could help here. Can you please eloborate?

> Maybe I missed it, but do we already bind a concrete GModelState class so that handlers can inject their concrete class directly?

Yes we do: https://github.com/eclipse-glsp/glsp-server/blob/0a51fc6aec435dd13d4b9a19f24ab6d8f4b77a1d/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/di/DiagramModule.java#L169

planger:
> > I haven't really made up my mind yet, but we could even go as far as injecting also the action / operation with the concrete type. This could make several abstract classes superfluous. But it would also make the interface of handlers a bit implicit. WDYT?
> 
> I'm a bit lost here. I don't see how injecting the action/operation could help here. Can you please eloborate?

I was wondering whether we could, before we invoke the handler, put the action (to be processed) with the concrete type into the DI container, so that handlers could also use `@Inject protected MyAction myAction;`. Then they wouldn't need to implement a generic typed interface.
One step further, we could to even allow them to not implement an interface at all but annotate the method that they want to be invoked on action processing and use method injection in this method.
However, this would also affect the registration of the action handler, as the interface doesn't show the action that they are repsonsible for.

> 
> > Maybe I missed it, but do we already bind a concrete GModelState class so that handlers can inject their concrete class directly?
> 
> Yes we do:
> 
> https://github.com/eclipse-glsp/glsp-server/blob/0a51fc6aec435dd13d4b9a19f24ab6d8f4b77a1d/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/di/DiagramModule.java#L169

Yes, but can clients now also inject a subclass of `GModelState` with `@Inject protected MyModelState myModelState;`

PR #138, opened 2021-10-31
#422 Cleanup ActionDispatcher API

tortmayr:
Remove dispatch method for action messages from API. Action dispatchers are now client session specific and do not need the full message context anymore. The action object alone is enough.

Part of eclipse-glsp/glsp/issues/422


planger (approved):

PR #140, opened 2021-11-02
#441 Always apply size in `ChangeBoundsOperationHandler`

tortmayr:


CamilleLetavernier (approved):
Looks good to me :+1:. With this, all layouts can decide which parameter they want to use (prefSize or size)

PR #141, opened 2021-11-03
#448 Implement dispatching of actions on next model-update

tortmayr:
and update `WorkflowLayoutEngine` to dispatch a `CenterAction` after the next model update.
Fixes https://github.com/eclipse-glsp/glsp/issues/448

### DIFF
--- a/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/layout/WorkflowLayoutEngine.java
+++ b/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/layout/WorkflowLayoutEngine.java
@@ -18,6 +18,8 @@ package org.eclipse.glsp.example.workflow.layout;
 import org.eclipse.glsp.graph.GGraph;
 import org.eclipse.glsp.layout.ElkLayoutEngine;
 import org.eclipse.glsp.layout.GLSPLayoutConfigurator;
+import org.eclipse.glsp.server.actions.ActionDispatcher;
+import org.eclipse.glsp.server.actions.CenterAction;
 import org.eclipse.glsp.server.model.GModelState;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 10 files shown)


examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/layout/WorkflowLayoutEngine.java
```diff
@@ -27,12 +29,16 @@
...
+   @Inject
+   protected ActionDispatcher actionDispatcher;
+
    @Override
    public void layout() {
       if (modelState.getRoot() instanceof GGraph) {
          GLSPLayoutConfigurator configurator = new GLSPLayoutConfigurator();
          configurator.configureByType("graph");
          this.layout((GGraph) modelState.getRoot(), configurator);
+         actionDispatcher.dispatchAfterNextUpdate(new CenterAction());
```
planger, line 41: I wouldn't expect that the layout engine queues an action after layouting -- so maybe this isn't the best example use case for `dispatchAfterNextUpdate`.

planger (approved):
Thanks, looks good to me!
However, I'm not sure the layout engine is in general the best example usage of `dispatchAfterNextUpdate`. At least to me this is quite surprising.
Other than that, this looks great!

examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/layout/WorkflowLayoutEngine.java
tortmayr: I personally think it's not an uncommon usecase to want  to trigger a Center action after (manually) autolayouting the model. However, I happily change this to a better example. Any suggestions?

ivy-cst: An Idea could be: Select a node after the creation operation.

planger: The use case to center after layout makes sense, but I find it a bit odd that the layout engine invokes this action. I'd rather expect the center action to be invoked from the layout handler. But I'm fine keeping it this way.
I like the idea by @ivy-cst though.

tortmayr: I also like the idea by @ivy-cst and updated the PR accordingly.

planger:
There is a CI error, other than that, it looks good! Thanks!

PR #142, opened 2021-11-09
 #432#257 Align server action with client protocol

tortmayr:
- Ensure that all request/response actions are implemented as proper request-response action pair
- Ensure that getter for optional action properties returns an Optional
- Ensure that action properties that have a collection type are always initialized
- Rename static action kind constant to `KIND`

Closes eclipse-glsp/glsp/issues/432
Part of eclipse-glsp/glsp/issues/257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,41 @@
+# Eclipse GLSP Server Changelog
+
+## [v0.9.0- 09/12/2021](https://github.com/eclipse-glsp/glsp/releases/tag/0.9.0)
+
+### Changes
+
+- [websocket] Fixed issue that was caused by reusing a shared injector for each client connection [#149](https://github.com/eclipse-glsp/glsp-server/pull/91)
+- [navigation] Added a utility class for 'JsonOpenerOptions' [#153](https://github.com/eclipse-glsp/glsp-server/pulls?q=92+93)
+- [diagram] Merged the `ServerLayoutConfiguration` API into the `DiagramConfiguration` API. [#123](https://github.com/eclipse-glsp/glsp-server/pull/95)
+- [model] Reworked model update flow. All aspects of the update process are now handled by the `ModelSubmissionHandler` [#122](https://github.com/eclipse-glsp/glsp-server/pull/95)
+- [model] Reworked and split `ModelFactory` API into a dedicated component responsible for loading the model source (`ModelSourceLoader`) and a component responsible for transforming the current modelstate into its GModel-representation (`GModelFactory`) [#119](https://github.com/eclipse-glsp/glsp-server/pull/96)
+- [protocol] Added implementation for `SetViewportAction` [#179](https://github.com/eclipse-glsp/glsp-server/pull/99)
+- [model] Introduced `GArgumentable` interface which is implemented by all `GModelElement`s. This can be used to provide additional information in form of an `args` map without having to extend the `GModel` type [#194](https://github.com/eclipse-glsp/glsp-server/pull/100)
+- [model] Extended default type mapping and added builder for `GArgumentable` elements and corresponding utility classes [#180](https://github.com/eclipse-glsp/glsp-server/pull/105)
+- [protocol] Added `fileUri` property to `SaveModelAction` and updated `SaveModelActionHandler` accordingly [#208](https://github.com/eclipse-glsp/glsp-server/pull/103/)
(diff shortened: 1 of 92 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/SetMarkersAction.java
```diff
@@ -15,22 +15,23 @@
...
 package org.eclipse.glsp.server.features.validation;
 
+import java.util.ArrayList;
 import java.util.List;
 
-import org.eclipse.glsp.server.actions.Action;
+import org.eclipse.glsp.server.actions.ResponseAction;
 
-public class SetMarkersAction extends Action {
+public class SetMarkersAction extends ResponseAction {
```
martin-fleck-at, line 23: Since you changed that into a response action I was worried that it might not properly work anymore to simply send that action from the server to the client without a request. However, as far as I can tell this still works since if there is no request for that the response id will be not set and the action is handled normally.

While I tested this I noticed that in `ResponseAction.respond` we do not filter for the respective reponse class, so if we have a request with id '1' and we return a list of responses they all get assigned the response id '1'. I think that may cause problems on the client where we wait for only one response for a request, so it may make sense to only assign the response id for one of the actions, what do you think? I believe something like this may work:

```
   public static Action respond(final Action request, final Action response) {
      if (request instanceof RequestAction<?>) {
         Class<?> responseClass = GenericsUtil.getGenericTypeParameterClass(request.getClass(), RequestAction.class);
         if (responseClass.isInstance(response)) {
            ((ResponseAction) response).setResponseId(((RequestAction<?>) request).getRequestId());
         }
      }
      return response;
   }
```

martin-fleck-at (changes requested):

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/SetMarkersAction.java
tortmayr: Fixed

tortmayr:
@martin-fleck-at I already included the Metadata changes for the upcoming release in this commit (versions, md updates, changelogs etc.). Please review them as well. Thanks!

CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
+# Eclipse GLSP Server Change Log
+
+## [v0.9.0- 20/10/2020](https://github.com/eclipse-glsp/glsp/releases/tag/0.9.0)
+
+[Server:](#server_changes_1.9.0)
```
martin-fleck-at, line 5: Should probably be called 'Changes:' since it is the server change log already.

CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
...
+- [model] Renamed `ModelFactory` to `ModelSourceLoader` and adapted interface method. This also affects implementing classes like the `JsonFileModelFactory` [#119](https://github.com/eclipse-glsp/glsp-server/pull/96)
+- [model] Reworked `ModelSubmissionHandler` API. This includes changes of method names and parameters [#119](https://github.com/eclipse-glsp/glsp-server/pull/96) [#197](https://github.com/eclipse-glsp/glsp-server/pull/101)
+- [operation] Refactored the `CreateNodeOperationHandler.createNode` method. The method now also passes the argument map of the operation. [#223](https://github.com/eclipse-glsp/glsp-server/pull/108)
+- [diagram] Refactored methods in`DiagramConfiguration` to correctly reflect `ShapeTypeHint` instead of `NodeTypeHint`.
+- [build] Update dependencies of Google Guava (>= 30.1) and Google Guice (>= 5.0.0). [#260](https://github.com/eclipse-glsp/glsp-server/pull/119)
+- [build] The `org.apache.log4j` dependency is not reexported any more when consumed as p2 artifact. You need to import `org.apache.log4j` in your own plugins if you use the log4j API there. [#430](https://github.com/eclipse-glsp/glsp/issues/430)
```
martin-fleck-at, line 33: This and a few other entries are present in the breaking changes and the regular changes section. I suggest to either only have changes in one of the two sections or to put ALL breaking changes in the regular changes section as well for consistency.

martin-fleck-at (approved):
Changes look good to me and everything still works as expected. Thank you for the great work, Tobias! I just have a few minor comments for the new change log, so maybe you can have a look at that but they should not prevent merging and may also be done in a follow-up PR.

PR #143, opened 2021-12-07
#460: Notify source/target feature change on source/target id change

martin-fleck-at:
Fixes https://github.com/eclipse-glsp/glsp/issues/460

Before:
glsp_delete_edge

After:
glsp_delete_edge_fix


tortmayr (approved):
Thanks Martin. Clean fix! LGTM :+1:

PR #144, opened 2021-12-07
#467: Use graph as default root element instead of plain model root

martin-fleck-at:
Fixes https://github.com/eclipse-glsp/glsp/issues/467

glsp_wrong_root_fix


tortmayr (approved):
Thanks Martin. Good catch! :+1:

PR #148, opened 2022-01-07
Improve styling and simplify workflow example

planger:
https://github.com/eclipse-glsp/glsp/issues/492

### DIFF
--- a/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/utils/WorkflowBuilder.java
+++ b/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/utils/WorkflowBuilder.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019-2021 EclipseSource and others.
+ * Copyright (c) 2019-2022 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
--- a/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/utils/WorkflowBuilder.java
+++ b/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/utils/WorkflowBuilder.java
@@ -31,9 +31,7 @@ import org.eclipse.glsp.graph.builder.AbstractGEdgeBuilder;
 import org.eclipse.glsp.graph.builder.AbstractGNodeBuilder;
 import org.eclipse.glsp.graph.builder.impl.GCompartmentBuilder;
 import org.eclipse.glsp.graph.builder.impl.GLabelBuilder;
-import org.eclipse.glsp.graph.builder.impl.GLayoutOptions;
 import org.eclipse.glsp.graph.util.GConstants;
-import org.eclipse.glsp.graph.util.GConstants.HAlign;
 
 public final class WorkflowBuilder {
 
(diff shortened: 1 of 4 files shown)


examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/utils/WorkflowBuilder.java
```diff
@@ -126,20 +124,14 @@ public void setProperties(final TaskNode taskNode) {
          taskNode.setName(name);
          taskNode.setTaskType(taskType);
          taskNode.setDuration(duration);
-         taskNode.setLayout(GConstants.Layout.VBOX);
-         taskNode.getChildren().add(createCompartment(taskNode));
```
ndoschek, line 130: Due to the change of the task node layout, the validation of task nodes regarding the first character to be upper-case (i.e. [`validateTaskNode_labelStartsUpperCase`](https://github.com/eclipse-glsp/glsp-server/blob/b4dbb5796fed14de0f33b6e3da5168b9247fefa5/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/marker/WorkflowModelValidator.java#L84) does not work as expected anymore. Could you please adapt this validation method?

ndoschek (commented):
Thanks Philip!
While testing the workflow example, I recognized that two validation features do not work anymore for the Workflow example:

The label edit validation should prompt `Name should be unique` while typing. Currently, this only works for the `Push` node, as here the `name` and label `text` value are identical. In general, the `name` and label `text` values differ for most task nodes in the example workflow model.
As I could observe, the label edit changes the label's text and the [validator checks the name of a task node](https://github.com/eclipse-glsp/glsp-server/blob/b4dbb5796fed14de0f33b6e3da5168b9247fefa5/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/labeledit/WorkflowLabelEditValidator.java#L43) instead. Could we align the behavior to get the correct label edit validation?

examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/utils/WorkflowBuilder.java
planger: Nice catch! Fixed with 9c1d1e3

ndoschek (changes requested):
Thanks for the update Philip! The label edit validation works now as expected! :tada: 

Unfortunately I didn't elaborate my [comment](https://github.com/eclipse-glsp/glsp-server/pull/148#discussion_r781180788) enough.
The validation in the `WorkflowModelValidator` (if a task name starts with an upper case letter) is also broken, because there is no dedicated compartment in the taskNode anymore (see [`validateTaskNode_labelStartsUpperCase`](https://github.com/eclipse-glsp/glsp-server/blob/b4dbb5796fed14de0f33b6e3da5168b9247fefa5/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/marker/WorkflowModelValidator.java#L84) ).

Could you have another look? Thanks!

planger:
Thanks for the great catch! I fixed the workflow validator in 453b347.

ndoschek (changes requested):
Great, thank you Philip for the quick fix! Looks great! :+1: 

I would have one final minor request, I just realized the file headers should be updated to `-2022`. Thanks!

planger:
Thanks, I updated the headers.

ndoschek (approved):
Thanks Philip! :tada:

PR #151, opened 2022-03-05
#571 Properly dispose temporary configuration session injectors

### LINKED ISSUE #571: Global ActionDispatcher Thread is started for each connection and will never be stopped
For each connection (over LSP or WebSocket) a global ActionDispatcher Thread is started in addition to the session. So There are two ActionDispatcher Threads for each connection and only the one on the session is stopped if the connection is closed.

I think, the global ActionDispatcher Thread is created over the `DefaultServerGsonConfigurator`.
```java
diagramModules.values().stream()
   .map(module -> serverInjector.createChildInjector(module).getInstance(ServerConfigurationContribution.class))
   .forEach(contribution -> {
       contribution.configure(actionRegistry);
       contribution.configure(graphGsonConfigurator);
    });
```

Also other classes are instantiated twice e.g the `ModelSourceWatcher`. The first time with a fall back client id and the second time with the correct one.

So I think all this classes are only used by the client session injector and don't needed to be started via the global server injector.

Screenshot:
Only one theia integration started (with some full page refreshes). Only every second thread is stopped.
image

tortmayr:
Refactor the  initial configuration of the `DefaultServerGsonConfigurator` to actually create temporary client sessions to retrieve the corresponding `ServerConfigurationContribution`. This session can then be safely disposed once its no longer needed (i.e. the configuration is completed).
By disposing the client session we ensure that all temporary created resources like the `ActionDispatcher` (thread) are properly disposed.
In addition, the acutal configuration logic is refactored into a protected method that is called from the constructor. This makes it easier for subclasses to customize the configuration behavior.

Fixes eclipse-glsp/glsp/issues/571

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/gson/DefaultServerGsonConfigurator.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/gson/DefaultServerGsonConfigurator.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2021 EclipseSource and others.
+ * Copyright (c) 2022 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


ivy-lli (approved):
Thanks for the fix

martin-fleck-at (approved):
Looks good to me! Thank you Tobias!

PR #153, opened 2022-03-22
Ensure that LayoutEngine is only injected into an optional field

planger:
Otherwise Guice would through an error if no layout engine is bound by
the module (which is fine however).

Fixes https://github.com/eclipse-glsp/glsp/issues/600


tortmayr (approved):
LGTM! Thanks for the fast fix.

PR #154, opened 2022-03-25
Refactorings regarding source model loading and saving

planger:
With this change, I'd like to consistently use the term source model and
avoid the term model source. This is to avoid the overlap with Sprotty's
concept of ModelSource and its unclear relation to a source model and
GModel.

I suggest to avoid Model Source altogether and instead introduce and
consistently use "source model" to refer to the underlying model that we
read and write. Thus we have in GLSP two types of models:
  * Source Model
  * Graph Model (GModel)

Thus, I did the following refactorings:
  * Rename ModelSourceLoader to SourceModelPersistence
  * Add saving to the SourceModelPersistence interface
  * Delegate from the SaveModelActionHandler to that implementation
  * Rename ModelSourceWatcher into SourceModelWatcher
  * Update javadoc and variable names where fit

Fixes https://github.com/eclipse-glsp/glsp/issues/582

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/JsonFileGModelStore.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/JsonFileGModelStore.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019-2021 EclipseSource and others.
+ * Copyright (c) 2019-2022 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -19,10 +19,14 @@ import static org.eclipse.glsp.server.types.GLSPServerException.getOrThrow;
 
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
 import java.io.Reader;
+import java.io.Writer;
 import java.nio.charset.StandardCharsets;
+import java.util.Map;
 import java.util.Optional;
 
 import org.apache.log4j.Logger;
@@ -30,6 +34,7 @@ import org.eclipse.glsp.graph.DefaultTypes;
 import org.eclipse.glsp.graph.GGraph;
 import org.eclipse.glsp.graph.GModelRoot;
 import org.eclipse.glsp.graph.GraphFactory;
+import org.eclipse.glsp.server.actions.SaveModelAction;
 import org.eclipse.glsp.server.gson.GraphGsonConfigurationFactory;
 import org.eclipse.glsp.server.model.GModelState;
 import org.eclipse.glsp.server.types.GLSPServerException;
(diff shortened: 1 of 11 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/JsonFileGModelStore.java
```diff
@@ -39,36 +44,31 @@
...
 
-   protected File convertToFile(final GModelState modelState) {
-      return getOrThrow(ClientOptionsUtil.getSourceUriAsFile(modelState.getClientOptions()),
-         "Invalid file URI:" + ClientOptionsUtil.getSourceUri(modelState.getClientOptions()));
-   }
-
    protected Optional<GModelRoot> loadSourceModel(final File file, final GModelState modelState) {
       try (Reader reader = new InputStreamReader(new FileInputStream(file), StandardCharsets.UTF_8)) {
-         Gson gson = gsonConfiguratior.configureGson().create();
+         Gson gson = gsonConfigurator.configureGson().create();
```
tortmayr, line 71: I'm wondering whether it's really necessary to construct a new Gson instance for every load/save request.
With method injection we could create a shared instance during injection time: 
```java
protected Gson gson;
@Inject
public void configureGson(GraphGsonConfiguratorFactory gsonConfigurator) {
gson =  gsonConfigurator.configureGson().setPrettyPrinting().create();
}
```

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/RequestModelActionHandler.java
```diff
@@ -21,7 +21,7 @@
 import org.eclipse.glsp.server.actions.AbstractActionHandler;
```
tortmayr, line 21: Please update the copyright header to -2022

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/SourceModelStorage.java
```diff
@@ -15,27 +15,39 @@
  ********************************************************************************/
```
tortmayr, line 15: Please update the copyright header to -2022

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/SourceModelStorage.java
```diff
@@ -15,27 +15,39 @@
...
     *
     * @param action Action sent by the client to specifying the information needed to load the source model.
     */
    void loadSourceModel(RequestModelAction action);
+
+   /**
+    * Saves the source model from the <code>modelState</code>.
+    *
+    * @param action Action sent by the client to specifying the information needed to save the source model.
+    * @return Returns <code>true</code> if save is done, <code>false</code> otherwise.
```
tortmayr, line 50: Technically the `GModelJsonStorage` implementation does not conform to this convention.
It either returns `true` or throws a `GLSPServerException`. `false` is never returned.
I'm wondering wether a boolean return value makes sense in this case.
I'd opt for `void` return and document that the method is expected to thrown an excpetion if something went wrong during the save process.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/FileWatcher.java
```diff
@@ -44,7 +44,7 @@
```
tortmayr, line 44: Please update the copyright header to -2022

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/modelsourcewatcher/SourceModelWatcher.java
```diff
@@ -16,18 +16,17 @@
 package org.eclipse.glsp.server.features.modelsourcewatcher;
```
tortmayr, line 16: Please update the copyright header to -2022

tortmayr (changes requested):
Thanks! In general the code looks good to me I have a few minor netpicks
and suggestions.
(Also I only did a code review because I  haven't had time yet to test the changes)

tortmayr:
This is the first breaking change since the 0.9.0 so maybe it's also a good idea to update the [changelog](https://github.com/eclipse-glsp/glsp-server/blob/master/CHANGELOG.md)

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/SourceModelStorage.java
planger: Agreed, I updated the code and javadoc accordingly.

tortmayr (approved):
LGTM! :+1:

PR #156, opened 2022-04-05
Improve workflow layouter

planger:
* Remove bendpoints, because they are inconvenient for editing later on
* Set alignmenet of node placement to balanced

### DIFF
--- a/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/layout/WorkflowLayoutEngine.java
+++ b/examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/layout/WorkflowLayoutEngine.java
@@ -15,6 +15,9 @@
  ********************************************************************************/
 package org.eclipse.glsp.example.workflow.layout;
 
+import org.eclipse.elk.alg.layered.options.FixedAlignment;
+import org.eclipse.elk.alg.layered.options.LayeredOptions;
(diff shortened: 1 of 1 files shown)


examples/org.eclipse.glsp.example.workflow/src/org/eclipse/glsp/example/workflow/layout/WorkflowLayoutEngine.java
```diff
@@ -31,8 +34,17 @@
    public void layout() {
       if (modelState.getRoot() instanceof GGraph) {
          GLSPLayoutConfigurator configurator = new GLSPLayoutConfigurator();
-         configurator.configureByType("graph");
-         this.layout((GGraph) modelState.getRoot(), configurator);
+         configurator.configureByType("graph")
+            .setProperty(LayeredOptions.NODE_PLACEMENT_BK_FIXED_ALIGNMENT,
```
tortmayr, line 38: I would also set the `EDGE_ROUTING` property to `polyline`. Otherwise we create a lot of unnecessary bendpoints that we have to remove afterwards.
```suggestion
            .setProperty(LayeredOptions.EDGE_ROUTING, EdgeRouting.POLYLINE),
            .setProperty(LayeredOptions.NODE_PLACEMENT_BK_FIXED_ALIGNMENT,
```

tortmayr (approved):
Looks good to me! The autolayouted diagram looks way better now. :+1:

PR #159, opened 2022-04-27
#622 Init org.eclipse.glsp.server.emf and add example models

ndoschek:
- Add project org.eclipse.glsp.server.emf
  - Add reusable EMF specific base classes
  - Add notation model (a copy of org.eclipse.emfcloud.modelserver.glsp.integration)
- Add project org.eclipse.glsp.example.tasklist
   - Add TaskList example model (a copy of org.eclipse.emfcloud.modelserver.glsp.integration)

Part of https://github.com/eclipse-glsp/glsp/issues/622

Contributed on behalf of STMicroelectronics.


ivy-cst:
Cool thanks :+1: . ModelState#updateRoot was the method we were missing. So we always had to removed all elements before syncing again from our model. Because with setRoot we otherwise lost the Command Stack.

PR #163, opened 2022-05-04
#579 Update to log4j 2.17.1

ndoschek:
- Update to org.apache.logging.log4j version 2.17.1
  - Migrate dependencies to log4j2
  - Migrate usage of logger to log4j2 and align naming to LOGGER
  - Migrate log configuration and introduce rolling log file policy
- Update target platform to latest 2022-03 release
- Adapt fatjar build
  - Setup fatjar to use multi-release jars to overcome logging issues while jar execution
  - Fix warnings related to overlapping classes during fatjar generation
- Ensure logging is configured for all server use cases on startup
  - Provide start methods to allow easier log configuration
- Add Log4j2Logger redirection for jetty server in case of websocket server usage
  - This is a copy from the org.apache.logging.log4j/log4j-appserver module, as this is not available as plugin.
- Improve help text for CLI usage
  - Add short program arguments for single letter argument usage
- Fix fragment host version for org.eclipse.glsp.graph.test in manifest file
- Update README

Resolves https://github.com/eclipse-glsp/glsp/issues/579


ivy-cst:
Cool that the newest libraries are used with this change. Just one thing. I relaxed that glsp depends on log4j with `Import-Package` instead of `Require-Bundle` so that the log4j classes can also come form another bundle than `org.apache.log4j`. See eclipse-glsp/glsp/issues/430. Would be nice if you could also depend on log4j2 with just `Import-Package`. Thanks.

ndoschek:
Thanks @ivy-cst! I updated my changes and import the packages now instead of requiring the bundle.

planger (approved):
Thanks @ndoschek, looks good to me!
I tested building and running the fatjar. That worked fine. :+1:

ndoschek:
Thanks @planger , yes I agree - I updated your suggested change

PR #165, opened 2022-05-13
Refactor classes that are only useful for GModel source models

planger:
Move base diagram module and operation handlers that operate directly on GModels (as a model source) to the dedicated package `org.eclipse.glsp.server.gmodel` and add prefix `GModel` in the class name.


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ModelSubmissionHandler.java
```diff
@@ -36,7 +36,7 @@
 @Singleton
 public class ModelSubmissionHandler {
 
-   @Inject()
+   @Inject
```
tortmayr, line 39: I'm in favor of removing unnecessary empty `()` from inject calls. But if we do that it should be done consistently.
I did a quick search and it seems like we are using `inject()` in multiple other places as well

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/gmodel/GModelDiagramModule.java
```diff
@@ -13,32 +13,39 @@
...
+ * serialization. They don't use any other information source to produce the GModel (see {@link GModelFactory}) and all
+ * operation handlers directly manipulate the GModel (as this is the source model in this scenario).
+ * </p>
+ * <p>
+ * This module is <b>not</b> intended to be used for editors that operate on another source model, such as an EMF model,
+ * etc. In those scenarios, a dedicated {@link SourceModelStorage}, {@link GModelFactory}, and set of operation
+ * handlers, etc. are required that load, transform (into a GModel) and manipulate that particular source model
+ * directly and re-create the GModel after each manipulation of the source model.
+ * For those scenarios please extend the {@link DiagramModule} directly or use a dedicated base module, such as
+ * the org.eclipse.glsp.server.emf.EMFDiagramModule.
```
tortmayr, line 41:
```suggestion
 * the `org.eclipse.glsp.server.emf.EMFDiagramModule`.
```

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/gmodel/GModelDiagramModule.java
```diff
@@ -49,20 +56,20 @@ protected Class<? extends GModelFactory> bindGModelFactory() {
...
-      binding.add(ChangeRoutingPointsHandler.class);
+      binding.add(GModelApplyLabelEditOperationHandler.class);
+      binding.add(GModelChangeBoundsOperationHandler.class);
+      binding.add(GModelChangeRoutingPointsHandler.class);
+      binding.add(GModelDeleteOperationHandler.class);
+      binding.add(GModelReconnectEdgeOperationHandler.class);
+      binding.add(GModelPasteOperationHandler.class);
       binding.add(CutOperationHandler.class);
-      binding.add(DeleteOperationHandler.class);
       binding.add(LayoutOperationHandler.class);
```
tortmayr, line 72: With the `LayoutOperationHandler` being generic and reusable for all implementations shouldn't it be bound as part of the parent module?

Same is probably true for the `CutOperationHandler`

tortmayr (changes requested):
Thanks @planger. Changes look mostly good to me, I only have a few inline comments.

One additional discussion point:

Due to our current naming scheme the handler names tend to become relatively 
long (especially implementation specific handlers for the emf/direct gmodel usecases). 
I'm wondering whether we should shorten this a little bit by dropping the `Action`/`Operation` part. At the moment action and operation names are unique anyways so appending this information adds no additional value.
e.g.  `GModelApplyLabelEditOperationHandler` => `GModelApplyLabelEditHandler`. 
@planger WDYT?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/SaveModelActionHandler.java
```diff
@@ -18,12 +18,17 @@
 import java.util.List;
 import java.util.Optional;
 
+import javax.inject.Inject;
```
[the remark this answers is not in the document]
planger: In most places we use google.inject. So I used this one now consistently.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/gmodel/GModelDiagramModule.java
planger: in fact, `LayoutOperationHandler` was already bound in the superclass anyway. But I now also pushed the cut operation handler up into the super class (and removed both from the subclass).

tortmayr (approved):
Thanks for the fast update. LGTM!

PR #172, opened 2022-06-02
#635: Execute all pending actions before disposing action dispatcher

martin-fleck-at:
Client session listeners only get notified about the session disposal after the session and its action dispatcher are disposed. As they might perform some cleanup we should ensure that all pending actions are executed before the action dispatcher is disposed and the listeners are notified.

Fixes https://github.com/eclipse-glsp/glsp/issues/635


tortmayr (approved):
Thanks Martin. Code looks good to me.

CamilleLetavernier (approved):
Looks good to me :+1:

PR #176, opened 2022-07-06
#622 Add glsp.server.emf source to p2 plugin

ndoschek:
Follow-up of https://github.com/eclipse-glsp/glsp/issues/622
- Add `glsp.server.emf.source` to p2 plugin to improve development experience

### DIFF
--- a/releng/org.eclipse.glsp.repository/category.xml
+++ b/releng/org.eclipse.glsp.repository/category.xml
@@ -27,6 +27,9 @@
    <bundle id="org.eclipse.glsp.server.emf">
       <category name="org.eclipse.glsp"/>
    </bundle>
+   <bundle id="org.eclipse.glsp.server.emf.source">
+      <category name="org.eclipse.glsp"/>
+   </bundle>
    <category-def name="org.eclipse.glsp" label="Eclipse GLSP"/>
    <category-def name="org.eclipse.glsp.examples" label="GLSP Examples"/>
 </site>


plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFSourceModelStorage.java
```diff
@@ -57,7 +57,7 @@ public void loadSourceModel(final RequestModelAction action) {
       URI resourceURI = URI.createFileURI(sourceURI);
 
       EditingDomain editingDomain = getOrCreateEditingDomain();
-      doLoadSourceModel(editingDomain.getResourceSet(), resourceURI, action);
+      doLoadSourceModel(editingDomain.getResourceSet(), resourceURI);
```
tortmayr, line 60: Since the 1.0.0 Release we have to be more careful unnecessary API breaks. (Every API break  results (in theory) in a  new major version release).

We should keep the additional action argument as some implementations may want to retrieve additional data from  the action options.

plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/notation/EMFNotationSourceModelStorage.java
```diff
@@ -36,14 +35,12 @@ protected ResourceSet setupResourceSet(final ResourceSet resourceSet) {
    }
 
    @Override
-   protected void doLoadSourceModel(final ResourceSet resourceSet, final URI sourceURI,
-      final RequestModelAction action) {
-      loadSemanticModel(resourceSet, sourceURI, action);
-      loadNotationModel(resourceSet, sourceURI, action);
+   protected void doLoadSourceModel(final ResourceSet resourceSet, final URI sourceURI) {
```
tortmayr, line 38: Same as above. Please avoid the API break and keep the action arg

plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/notation/EMFNotationSourceModelStorage.java
```diff
@@ -52,8 +49,7 @@ protected URI deriveNotationModelURI(final URI sourceURI) {
       return sourceURI.trimFileExtension().appendFileExtension(NotationUtil.DEFAULT_EXTENSION);
    }
 
-   protected void loadSemanticModel(final ResourceSet resourceSet, final URI sourceURI,
-      final RequestModelAction action) {
+   protected void loadSemanticModel(final ResourceSet resourceSet, final URI sourceURI) {
       loadResource(resourceSet, deriveSemanticModelURI(sourceURI)).ifPresent(modelState::setSemanticModel);
```
tortmayr, line 53: Same as above. Please avoid the API break and keep the action arg

tortmayr (changes requested):
Thanks Nina. The actual P2 change looks good to me,
however the side change (removing the action arg from the `EMFSourceModelStorage`)is actually an API break and should therefore be avoided. In addition, implementations lose the access to the options of the `RequestModelAction`.

ndoschek:
Thanks @tortmayr! Agreed, I updated the PR!

tortmayr (approved):

PR #180, opened 2022-09-22
Update tycho maven version: 2.3.0 => 2.7.5

xai:
This is needed to build with maven and Java 17.
Related to this, includePackedArtifacts had to be disabled as it seems to cause issues with Java 14 and higher.

Closes https://github.com/eclipse-glsp/glsp/issues/750

Signed-off-by: Olaf Lessenich <olessenich@eclipsesource.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,7 @@
 			</activation>
 
 			<properties>
-				<tycho-version>2.3.0</tycho-version>
+				<tycho-version>2.7.5</tycho-version>
 				<maven-antrun-plugin.version>1.8</maven-antrun-plugin.version>
 			</properties>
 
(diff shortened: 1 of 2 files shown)


xai:
I need to have another look at this, there is a currently a mix of tycho-versions in the `pom.xml` files.

ndoschek (approved):
Thank you @xai, looks good to me, builds with both Java 11 and Java 17 and still works as expected :tada: 
Internally (e.g. for compiling) we still use Java 11, I will open up an issue for discussion to use Java 17 throughout the project in the future.

tortmayr:
> Internally (e.g. for compiling) we still use Java 11, I will open up an issue for discussion to use Java 17 throughout the project in the future.

I think we discussed this previously and decided that we in general want to support the last two LTS Versions of Java (see also https://endoflife.date/java). That's why the code is still compiling to 11. However, I'm open for discussing this 👍🏼 .

PR #183, opened 2022-11-11
781: Add RequestExportSvgAction and register it as default client action

ivy-lli:
Solve: https://github.com/eclipse-glsp/glsp/issues/781

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/RequestExportSvgAction.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/RequestExportSvgAction.java
@@ -1,0 +1,38 @@
+/********************************************************************************
+ * Copyright (c) 2019-2022 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+package org.eclipse.glsp.server.actions;
+
+/**
+ * Request the client to export an SVG from the current diagram.
+ */
+public class RequestExportSvgAction extends Action {
+
+   public static final String KIND = "requestExportSvg";
+
+   private String requestId;
+
+   public RequestExportSvgAction() {
+      super(KIND);
+   }
+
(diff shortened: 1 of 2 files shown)


planger:
Thank you for looking into the SVG export in Eclipse, @ivy-lli!

Your PR looks good to me in general.

However, do we only need this new action in the Eclipse integration? Afair, in Theia and VS Code, the menu item / command directly triggers an `ExportSvgAction` sent to the server, so there is no need for this action.
If this is true, I guess we have two options:
1. Either we add a `RequestExportSvgAction` to the server as this PR proposes, but then we'll have to add a handler to the other clients (VS Code and Theia) too, and add it to the [protocol documentation](https://github.com/eclipse-glsp/glsp/blob/master/PROTOCOL.md), as it then essentially becomes part of the protocol. This would enable the server to trigger an SVG export -- which may be a feature that's nice to have. However, then it needs to be supported for all client integrations.
2. We only add the `RequestExportSvgAction` to the Eclipse integration, indicating that it is an Eclipse-specific action only needed there, and configure everything that is needed in the Eclipse-specific [IdeServerModule](https://github.com/eclipse-glsp/glsp-eclipse-integration/blob/master/server/plugins/org.eclipse.glsp.ide.editor/src/org/eclipse/glsp/ide/editor/di/IdeServerModule.java). We already have something rather similar with the clipboard handling, for which Eclipse-specific actions are used.

I might be wrong, but I'd tend to prefer option 2 for now, unless there is a good use case for other integrations too.
WDYT?

ivy-lli:
Hi @planger 
There is a "Handler" for this in the general glsp-client, provided by sprotty: https://github.com/eclipse/sprotty/blob/master/packages/sprotty/src/features/export/export.ts#L56

And no I think the theia integration for example does not fire a `ExportSVGAction` but a `RequestExportSvgAction`: https://github.com/eclipse/sprotty-theia/blob/master/src/theia/diagram-commands.ts#L180 
And the same does the VSCode integration: https://github.com/eclipse-glsp/glsp-vscode-integration/blob/master/packages/vscode-integration/src/quickstart-components/configuration-util.ts#L60

And of course both of them have an own handler for the `ExportSVGAction`:

- https://github.com/eclipse-glsp/glsp-vscode-integration/blob/master/packages/vscode-integration/src/glsp-vscode-connector.ts#L394-L417
- https://github.com/eclipse-glsp/glsp-theia-integration/blob/master/packages/theia-integration/src/browser/diagram/glsp-theia-diagram-server.ts#L74-L77

But yes, maybe it's better to only register this action in the eclipse integration, as it maybe is no usecase that the server can trigger this action. I will change the [eclipse-integration PR](https://github.com/eclipse-glsp/glsp-eclipse-integration/pull/65) to this approach. Then would be this PR obsolete.

planger:
Thanks for clarifying! I forgot that this action is actually handled in Sprotty (just did a brief grep in the GLSP code).

> But yes, maybe it's better to only register this action in the eclipse integration, as it maybe is no usecase that the server can trigger this action. I will change the https://github.com/eclipse-glsp/glsp-eclipse-integration/pull/65 to this approach. Then would be this PR obsolete.

Ok great! I'll close this PR then and have a look at https://github.com/eclipse-glsp/glsp-eclipse-integration/pull/65

PR #184, opened 2022-12-13
Update Jenkinsconfiguration

tortmayr:
- directly include deploy steps instead of spawning a new dedicated
- jobs enable error reporting when master builds fails

Part of https://github.com/eclipse-glsp/glsp/issues/794


planger (approved):

PR #185, opened 2023-01-20
Update server.websocket to Jetty 11

tortmayr:
- Update default target platform
   - Eclipse 21.03-> 22.12
   - Jetty 9.x -> Jetty 11.0.12
   - LSP4J -> 0.8.0 -> 0.19.0  
   - ELK 0.7.0 -> 0.8.1
   - Add slf4j to enable logging in Jetty 10

- Consume log4j,slf4j and javax.servlet  via maven
   - Jetty websockt requires slf4j 2.0 which is not available via orbit and not compatible with the orbit log4j version. So we instead use the new target platform features and consume it directly from a maven location

-  Update server.websocket plugin to conform to Jetty API breaks

- Update to ELK 0.8.1 and remove workaround (Fixes https://github.com/eclipse-glsp/glsp/issues/265)
- Clean up pom files and consistently use properties for all versions
- Remove xtext nature from parent project. Otherwise we get false positives in the Eclipse Problem view.
- Ensure that both the p2 and m2 build use the same version ranges for their dependencies (if possible)

- Provide a second target for 2021-09  to use it as minimal baseline.

Part of https://github.com/eclipse-glsp/glsp/issues/764
Contributed on behalf of STMicroelectronics


tortmayr:
Note, in order to use maven locations in the target/tpd files you have to make sure that 
-the m2e PDE Integration feature is installed (available via the default eclipse update site)
- A recent nightly version of the TPD plugin is installed (https://download.eclipse.org/justj/?file=cbi/updates/tpd/nightly/latest)

planger (approved):
Excellent work, thank you! I've tested with 2022-12 and 2021-09 TPDs.

PR #186, opened 2023-01-23
Switch to Jetty 10

tortmayr:
- Using Jetty 11 causes unnecessary Version resolution conflicts with the Jetty Version used by Eclipse (10) =>
Align versions by using Jetty 10 for GLSP instead of 11
- Remove reexporting of jetty dependencies to avoid potential version conflicts
- Refactor glsp.graph package to avoid usage of internal API that is no longer exposed with Gson >=2.8.9

Contributed on behalf of STMicroelectronics


planger (approved):
Thank you, looks good to me and workflow example still works as expected!

PR #187, opened 2023-01-23
Let OperationHandlers return an optional command to control execution

martin-fleck-at:
- Elevate concept of command to OperationHandler for more control 
-- Move command handling from EMF to general operation handler

- Let abstract handlers use recording command for compatibility
-- Add new 'preExecute' method to abort execution

- Rewrite compound operation handler to create compound command

- Minor fixes
-- Convert ApplyTaskEditActionHandler from operation to action handler

Fixes https://github.com/eclipse-glsp/glsp/issues/864


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changes
 
 - [graph] Updated the Graph model and add Layoutable interface [#175](https://github.com/eclipse-glsp/glsp-server/pull/175) - Contributed on behalf of STMicroelectronics
+- [operation] Rework `OperationHandler` to provide an optional command instead of direct execution to allow more execution control [#187](https://github.com/eclipse-glsp/glsp-server/pull/187)
```
tortmayr, line 8: Should probably be declared as breaking change because the `OperationHandler` interface is no longer compatible.
Also it would be nice to mention that we deprecated the `AbstractXYZ` implementations and removed the old `BasicXYZOperationHandler`s (which where announced to be removed before 1.0.0, so we even kept them longer as necessary)

tortmayr (changes requested):
Thanks Martin! Create change and even nicer API refactoring that cleans up the convoluted OperationHandler type hierarchy without introducing too many API breaks.

Apart from a couple of inline suggestions I noticed the following:
Part of the example operation handlers still use the (now) deprecated handlers which results in a couple of warnings:
Screenshot from 2023-01-26 10-16-43ec6ac70cb0b6.png)

We should clean that up.

Also, since we are touching the `Operationhandler` API anyways, I was wondering wether we also should refactor and deprecate the `CreateEdgeOperationHandler` and `CreateNodeOperationHandler` sub interfaces and base implementations.
From what I can see, their only purpose right now is to enable code sharing of utility functions for the `CreateNode` use case. The `CreateEdgeOperationHandler` even is a completely empty marker interface. Maybe it would be better to move the utility functions in question into a util class and remove the unnecessary subtypes of `CreateOperationHandler` ?

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
        - Jetty 9.x -> Jetty 10.0.13
        - LSP4J -> 0.8.0 -> 0.11.0
        - ELK 0.7.0 -> 0.8.1
+- [operation] Rework `OperationHandler` to provide an optional command instead of direct execution to allow more execution control [#187](https://github.com/eclipse-glsp/glsp-server/pull/187)
+    - `Abstract<XYZ>` base implementations were deprecated in favor of `GModelOperationHandler` and `EMFOperationHandler` base classes
+    - Long-termin deprecated and unused `Basic<XYZ>` base classes were removed
```
tortmayr, line 17:
```suggestion
    - Long-term deprecated and unused `Basic<XYZ>` base classes were removed
```

tortmayr:
Awesome! Changes look good to me. 👍🏼

tortmayr (approved):

PR #189, opened 2023-01-26
don't wait on stdin in websocket server

mockersf:
closes https://github.com/eclipse-glsp/glsp/issues/883

remove listening on stdin, so that the server can be started without access to stdin and rely on process control to stop


tortmayr (approved):
Thanks, LGTM! It definitely makes sense to not rely on stdin here.

PR #191, opened 2023-01-30
Remove unnecessary 'ApplyTaskEditAction' and update operation handlers

martin-fleck-at:
- Use 'EditTaskOperation' directly from the client
- Minor checkstyle cleanup and refactoring for Paste handler
- Make CutOperationHandler GModel-specific as it returns recording cmd
- Let LayoutOperationHandler inherit from generic base class
- Update some JavaDoc

Relates to https://github.com/eclipse-glsp/glsp/issues/864


tortmayr (approved):
LGTM! 👍🏼

PR #197, opened 2023-03-22
GLSP-964 Properly bind WorkflowGLSPServer

### LINKED ISSUE #964: Properly bind WorkflowGLSPServer
The `WorkflowGLSPServer` is intended as a demonstrator on how to handle custom arguments on server initialization.
However, currently this class is not bound in any module and therefore unused.
The Workflow example should properly bind and use this custom server.

tortmayr:
Fixes https://github.com/eclipse-glsp/glsp/issues/964

Also: Update copyright header for new files to use a dynamic year variable instead of  hardcoded "2020"  (org.eclipse.glsp.examples.workflow)


planger (approved):

PR #198, opened 2023-04-06
feat: adjust server startup to use autoassigned ports

xai:
This sets the default port to 0, which will cause the server to automatically assign a port. The port is then printed to the console in the startup message and can be parsed by clients.

Part of eclipse-glsp/glsp#965

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
   - EMFModelStateImpl and EMFNotationModelStateImpl classes have been added
   - Related Modules have been updated to inject these GModelState sub-types as a Singleton
 - [validation] Add explicit support and API for live and batch validation [#200](https://github.com/eclipse-glsp/glsp-server/pull/200)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 - [operation] Rework `OperationHandler` to provide an optional command instead of direct execution to allow more execution control [#187](https://github.com/eclipse-glsp/glsp-server/pull/187)
   - `Abstract<XYZ>` base implementations were deprecated in favor of `GModelOperationHandler` and `EMFOperationHandler` base classes
   - Long-term deprecated and unused `Basic<XYZ>` base classes were removed
+- [server] Default port has changed from 5007 to 0, which implies autoassignment by the OS
```
tortmayr, line 22: It don`t think it's necessary to split this into bullet points. One pullet points covering both should be enough. Please also adda link to the PR at the end like this: [#198](https://github.com/eclipse-glsp/glsp-server/pull/198/)

tortmayr (approved):
LGTM. Only one minor nitpick regarding the changelog

tortmayr (approved):
LGTM! Thanks

PR #199, opened 2023-04-12
GLSP-974: Use interface-injection for all subclasses of ModelState

CamilleLetavernier:
- Use interfaces for EMFModelState and EMFNotationModelState
- Use these interfaces for binding and injection

refs https://github.com/eclipse-glsp/glsp/issues/974

### DIFF
--- a/plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFModelState.java
+++ b/plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFModelState.java
@@ -3,11 +3,16 @@
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
- * https://www.eclipse.org/legal/epl-2.0, or the MIT License which is
- * available at https://opensource.org/licenses/MIT.
+ * https://www.eclipse.org/legal/epl-2.0.
  *
- * SPDX-License-Identifier: EPL-2.0 OR MIT
- ********************************************************************************/
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ******************************************************************************/
 package org.eclipse.glsp.server.emf;
 
 import org.eclipse.emf.ecore.resource.ResourceSet;
--- a/plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFModelState.java
+++ b/plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFModelState.java
@@ -15,7 +15,7 @@ import org.eclipse.emf.edit.domain.EditingDomain;
 import org.eclipse.glsp.server.model.GModelState;
 
 public interface EMFModelState extends GModelState {
-   void setEditingDomain(final EditingDomain editingDomain);
+   void setEditingDomain(EditingDomain editingDomain);
 
    EditingDomain getEditingDomain();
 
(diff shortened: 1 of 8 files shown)


planger:
Nice! Maybe is https://github.com/eclipse-glsp/glsp/issues/896 or is/could be addressed in this go?

CamilleLetavernier:
> Maybe is https://github.com/eclipse-glsp/glsp/issues/896 or is/could be addressed in this go?

I'll take a look :+1:

CamilleLetavernier:
I believe this is fixed by this PR, as we now bind the class only once. Extra bindings only specify additional interface-bindings for the same class. However I'm not really sure how to test it. Is there an example where I could try it?

planger:
> I believe this is fixed by this PR, as we now bind the class only once. Extra bindings only specify additional interface-bindings for the same class.

That's what I also was thinking looking at the code.

> However I'm not really sure how to test it. Is there an example where I could try it?

Well, I think it should be sufficient if you'd remove the override of bindGModelState() in your subclass of EMFDiagramModule. The use case was basically having a concrete DiagramModule that extends EMFDiagramModule but uses the EMFModelState instead of binding a specialization of EMFModelState (in its bindGModelState() override).

plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFModelStateImpl.java
```diff
@@ -0,0 +1,115 @@
+/********************************************************************************
+ * Copyright (c) 2022 EclipseSource and others.
```
ndoschek, line 2:
```suggestion
 * Copyright (c) 2022-2023 EclipseSource and others.
```

plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/notation/EMFNotationModelStateImpl.java
```diff
@@ -0,0 +1,87 @@
+/********************************************************************************
+ * Copyright (c) 2022 EclipseSource and others.
```
ndoschek, line 2:
```suggestion
 * Copyright (c) 2022-2023 EclipseSource and others.
```

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/model/DefaultGModelState.java
```diff
@@ -42,6 +44,10 @@ public class DefaultGModelState implements GModelState {
    protected String editMode;
    protected GModelIndex index = GModelIndex.empty();
 
+   public DefaultGModelState() {
+      System.err.println("Instantiate Model State " + getClass().getName());
```
ndoschek, line 48: I assume this line is a leftover?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/model/DefaultGModelState.java
ndoschek, line 1: Update year range in header.

ndoschek (commented):
Thanks a lot @CamilleLetavernier!
Changes look great to me, I did not really test it, but it will be easier to test it with the already opened [PR](https://github.com/eclipse-emfcloud/modelserver-glsp-integration/pull/53#event-8982714256) in the modelserver-glsp-integration.

Besides that I only have a few very minor comments, mostly regarding the year ranges in the file headers.

plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFModelState.java
```diff
@@ -1,110 +1,26 @@
 /********************************************************************************
- * Copyright (c) 2022 EclipseSource and others.
+ * Copyright (c) 2023 EclipseSource and others.
```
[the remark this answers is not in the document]
CamilleLetavernier: In this case it's a new interface (even though it replaces a class with the same file name), so I think it needs new copyright headers

ndoschek (approved):
Thanks @CamilleLetavernier for the updates! LGTM :+1:

CamilleLetavernier:
Commit a0350bf updates the Changelog for 2.x, as this PR includes breaking changes related to EMFModelState and EMFNotationModelState

ndoschek (approved):

PR #200, opened 2023-04-20
Add validation reason and distinguish between live and batch validation

planger:
https://github.com/eclipse-glsp/glsp/issues/980

For review, please see [the comment on which changes belong together](https://github.com/eclipse-glsp/glsp/issues/980#issuecomment-1516867263).

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/ModelValidator.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/ModelValidator.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019-2021 EclipseSource and others.
+ * Copyright (c) 2019-2023 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -15,6 +15,8 @@
  ********************************************************************************/
 package org.eclipse.glsp.server.features.validation;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import org.eclipse.glsp.graph.GModelElement;
@@ -23,6 +25,14 @@ import org.eclipse.glsp.graph.GModelElement;
  * Validates a list of {@link GModelElement}s based on a set of validation rules and returns corresponding issue
  * {@link Marker}s.
  * An issue marker is a serializable description of the validation violation that can be visualised by the GLSP client.
+ *
+ * There are two default reasons for a validation:
+ * <ol>
+ * <li><i>live</i> validation is executed on open and after each operation. These validation rules should be rather
+ * fast. Overwrite {@link #doLiveValidation(GModelElement)} to implement the live validation rules.</li>
+ * <li><i>batch</i> validation is executed on demand by the client. These validation rules can be more expensive.
+ * Overwrite {@link #doBatchValidation(GModelElement)} to implement the batch validation rules.</li>
+ * </ol>
  */
 public interface ModelValidator {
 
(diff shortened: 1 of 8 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/ModelValidator.java
```diff
@@ -23,14 +25,73 @@
...
     * @param elements The list of {@link GModelElement} to validate.
+    * @param reason   The reason for a validation request, such as "batch" or "live" validation.
     * @return A list of {@link Marker}s for the validated {@link GModelElement}s.
     */
-   List<Marker> validate(GModelElement... elements);
+   default List<Marker> validate(final List<GModelElement> elements, final String reason) {
+      List<Marker> markers = new ArrayList<>();
+
+      for (GModelElement element : elements) {
+         if (MarkersReason.LIVE.equals(reason)) {
```
tortmayr, line 50: What is the assumption here regarding the reason property? Is it a fixed string that can only either be `batch` or `live`? Or can adopters also introduce custom reasons? If so we should probably explicitly check for the `batch` reason as well and add a fallback case for unsupported reasons.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/ModelValidator.java
```diff
@@ -23,14 +25,73 @@
...
+            markers.addAll(doLiveValidation(element));
+         } else {
+            markers.addAll(doBatchValidation(element));
+         }
+         if (element.getChildren() != null) {
```
tortmayr, line 55: I think this is a unnecessary null-check. List properties of EObjects are always defined and should never be null. We should rather check if the children list is not empty.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/RequestMarkersHandler.java
```diff
@@ -43,27 +42,24 @@ public class RequestMarkersHandler extends AbstractActionHandler<RequestMarkersA
...
-      List<Marker> markers = new ArrayList<>();
-      GModelIndex currentModelIndex = modelState.getIndex();
-      for (String elementID : elementsIDs) {
-         Optional<GModelElement> modelElement = currentModelIndex.get(elementID);
-         if (modelElement.isPresent()) {
-            markers.addAll(validator.get().validate(modelElement.get()));
-         }
 
-      }
+      String reason = action.getReason() != null ? action.getReason() : MarkersReason.BATCH;
```
tortmayr, line 57: See comment above. If the default is already properly set in the action class we don't need to handle it explicitly here.

tortmayr (changes requested):
Thanks Philip, support for live validation is definitely a nice feature addition.
I did not have time to test the changes yet and only did a preliminary code review for now.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/validation/ModelValidator.java
planger: Right, the assumption is that we use batch or live by default, but adopters may add custom reasons, e.g. some specific analysis and use custom strings here. My thinking was that they then would have to overwrite the `validate()` method. But you are right, it'd be more convenient to handle the batch case explicitly and have a `default doValidationForCustomReason()` that needs to be overwritten in those cases.

planger:
Thank you for the excellent review! I've applied your suggestions with b8a020f.

tortmayr (approved):
Thanks for the fast update! LGTM 👍🏼

PR #205, opened 2023-07-14
Add support for progress reporting

planger:
https://github.com/eclipse-glsp/glsp/issues/1019

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/utils/ServerMessageUtil.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/utils/ServerMessageUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020-2022 EclipseSource and others.
+ * Copyright (c) 2020-2023 EclipseSource and others.
  *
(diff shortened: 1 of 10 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/utils/ServerMessageUtil.java
```diff
@@ -57,35 +57,6 @@ public static ServerMessageAction error(final Exception e) {
       return error(getMessage(e), getDetails(e));
    }
 
-   public static ServerMessageAction message(final Severity severity, final String message,
-      final int timeout) {
```
tortmayr, line 61: Any specific reason why we remove this entirely instead of just removing the timeout parameter?

planger: I think because they are already above, if I'm not mistaken?

tortmayr: Ah makes sense. I didn't look at the full scope of the file 😄

PR #210, opened 2023-09-11
211: Revise TypeHints and server side feedback for creation actions

CamilleLetavernier:
- Add a new parameter to EdgeTypeHint, 'dynamic', indicating that new
edges need to check with the server before allowing creation
 -  Make source/target element type ids in `EdgeTypeHint` optional
  If not defined, all potential element types are considered to be valid sources/targets
- Add `RequestEdgeCheckAction` and `EdgeCheckResultAction` response to implement
 the dynamic check
- Add a optional `EdegeCreationchecker` component that can be implemented by adopters to provide dynamic typehints
- Adapt workflow example to use dynamic edge hints for weighted edges
- Move typehints related components into feature subpackage
- Move progress actions into progress feature subpackage

Co-authored-by: Camille Letavernier <cletavernier@eclipsesource.com >


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/CheckEdgeTargetResultAction.java
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+package org.eclipse.glsp.server.actions;
+
+/**
+ * {@link ResponseAction} for a {@link RequestCheckEdgeTargetAction}.
+ */
+public class CheckEdgeTargetResultAction extends ResponseAction {
+
+   public static final String KIND = "setTargetTypeHints";
```
tortmayr, line 23:
```suggestion
   public static final String KIND = "checkEdgeTargetResult";
```
Nees to match the KIND defiend in the client protocol

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/diagram/DefaultRequestTargetTypeHintsActionHandler.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.google.inject.Inject;
+
+/**
+ * Default handler implementation for {@link RequestCheckEdgeTargetAction}.
+ * This handler always rejects creation of dynamic Edges. Client applications
+ * should replace this default implementation with their own logic to
+ * conditionally accept creation of new edges configured with a Dynamic type.
+ */
+public class DefaultRequestTargetTypeHintsActionHandler extends AbstractActionHandler<RequestCheckEdgeTargetAction> {
```
tortmayr, line 33: We typically omit the default prefix
```suggestion
public class RequestTargetTypeHintsActionHandler extends AbstractActionHandler<RequestCheckEdgeTargetAction> {
```

tortmayr (changes requested):
Thanks @CamilleLetavernier.
I only had time to do a preliminary code review for now, but I will also test the changes later today

CamilleLetavernier:
Thanks,
I've pushed a commit to address the issues. I'm surprised that the code still worked despite the consistent Action Type; but I guess it's because it's a Request/Response? We don't actually have an ActionHandler for the response, so I think the type is ignored on the client side.
Looking at `ResponseAction` on the server, I think I should be using `ResponseAction.repond()` instead of simply returning the response action?

tortmayr:
>Looking at ResponseAction on the server, I think I should be using ResponseAction.repond() instead of simply returning the response action?
No, that's fine. The `ActionDistpacher` has some default behavior in place to automatically call `ResponseAction.response` for the return actions of `ActionHandlers`.
https://github.com/eclipse-glsp/glsp-server/blob/a48db7534ae1a729a3141ae50080b9c05ceb7eb5/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/internal/actions/DefaultActionDispatcher.java#L193

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/diagram/RequestTargetTypeHintsActionHandler.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Default handler implementation for {@link RequestCheckEdgeTargetAction}.
+ * This handler always rejects creation of dynamic Edges. Client applications
+ * should replace this default implementation with their own logic to
+ * conditionally accept creation of new edges configured with a Dynamic type.
+ */
+public class RequestTargetTypeHintsActionHandler extends AbstractActionHandler<RequestCheckEdgeTargetAction> {
+
+   @Inject
+   protected DiagramConfiguration diagramConfiguration;
```
tortmayr, line 36: Why do we need the `diagramConfiguration` here? It is unused.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/types/EdgeTypeHint.java
```diff
@@ -25,12 +27,19 @@ public class EdgeTypeHint extends ElementTypeHint {
...
 
    public EdgeTypeHint() {}
 
    public EdgeTypeHint(final String elementTypeId, final boolean repositionable, final boolean deletable,
       final boolean routable,
       final List<String> sourceElementTypeIds, final List<String> targetElementTypeIds) {
+      this(elementTypeId, repositionable, deletable, routable, false, sourceElementTypeIds, targetElementTypeIds);
+   }
+
+   public EdgeTypeHint(final String elementTypeId, final boolean repositionable, final boolean deletable,
```
tortmayr, line 40: Not directly related to you change, but while we are at it we should probably make sure that the `createDefaultEdgeTypeHint` method of the `DiagramConfiguration` passes new, empty Lists for the source and target ids.
https://github.com/eclipse-glsp/glsp-server/blob/a48db7534ae1a729a3141ae50080b9c05ceb7eb5/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/diagram/DiagramConfiguration.java#L74

Otherwise the corresponding properties in the `SetTypeHintsAction` are not defined which can cause issues on the client

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/RequestCheckEdgeTargetAction.java
```diff
@@ -0,0 +1,69 @@
...
+import org.eclipse.glsp.server.types.EdgeTypeHint;
+
+/**
+ * A {@link RequestAction} used to check the validity of an edge being created. This is used
+ * to update creation feedback on the client side, for edges configured with a dynamic {@link EdgeTypeHint}.
+ *
+ * @see EdgeTypeHint#isDynamic()
+ * @see CheckEdgeTargetResultAction
+ */
+public class RequestCheckEdgeTargetAction extends RequestAction<CheckEdgeTargetResultAction> {
```
tortmayr, line 27: Do we really need the request prefix here?  IMO `CheckEdgeTargetAction` should be fine

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/CheckEdgeTargetResultAction.java
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+package org.eclipse.glsp.server.actions;
+
+/**
+ * {@link ResponseAction} for a {@link RequestCheckEdgeTargetAction}.
+ */
+public class CheckEdgeTargetResultAction extends ResponseAction {
+
+   public static final String KIND = "checkEdgeTargetResult";
```
tortmayr, line 23: Maybe we should also attach the type, source and target id here for additional context?
Otherwise adopters always would have to retrieve the corresponding request action for full context.

tortmayr (commented):
Thanks for addressing the issues @CamilleLetavernier. In general the change looks good to me.


However, currently it's not really possible to test this change without further modifications.
Maybe we could provide a simple custom `RequestTargetTypeHintsActionHandler` for the workflow example? For instance we could make the type hints for `weighted edges` dynamic and check on the server side whether the target is valid (I.e. source is a decision node, target is a task or fork/join node)

Also I'm not the biggest fan of having default no-op `RequestTargetTypeHintsActionHandler` that needs to be rebind by adopters. Maybe we could use a generic action handler that delegates to an optional injected service? Similar to how we handle e.g. `RequestModelActions`, `ReuestMarkersAction`, `LayoutOperation` etc.

We could introduce a `EdgeTargetValidator` service that can be bound by adopters, and the handler for the `RequestCheckEdgeTargetAction` simply delegates to the validator (if bound) 
@planger WDYT?

planger:
> We could introduce a EdgeTargetValidator service that can be bound by adopters, and the handler for the RequestCheckEdgeTargetAction simply delegates to the validator (if bound)
@planger WDYT?

Yes that would be better. Having to rebind a no-op is very implicit, being able to overwrite an empty method in the abstract class is more explicit. Our credo was always to have abstract methods for the mandatory implementations an adopter need to deliver, and the they can press Ctrl+space in their module and see what other optional implementations they can provide.

In any case, I would try to avoid the term "validator" and rather use "check" or so, to avoid confusing this concept with the model validation.

tortmayr:
@martin-fleck-at I have taken over this draft from Camille and have finalized the PR. Could you please have a look at it?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/typehints/EdegeCreationChecker.java
```diff
@@ -0,0 +1,49 @@
...
+ * A dynamic edge type hint is used for cases where a plain list of allowed source and target element ids is not enough
+ * to determine
+ * wether an edge beeing created is valid. In this cases the client will query the server to determine wether the edge
+ * is valid.
+ * The `EdegeCreationChecker` then checks the given edge information and returns wether the edge beeing created is
+ * valid.
+ * If no `EdgeTargetChecker` implementation is bound the creation of dynamic edge will always be rejected.
+ *
+ */
+public interface EdegeCreationChecker {
```
martin-fleck-at, line 33: Typo: `EdegeCreationChecker` -> `EdgeCreationChecker`

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/typehints/RequestCheckEdgeTargetActionHandler.java
```diff
@@ -0,0 +1,75 @@
...
+   protected GModelState modelState;
+
+   @Override
+   protected List<Action> executeAction(final RequestCheckEdgeAction action) {
+      return List.of(new CheckEdgeTargetResultAction(validate(action), action));
+   }
+
+   protected boolean validate(final RequestCheckEdgeAction action) {
+      if (!edgeTargetChecker.isPresent()) {
+         return false;
```
martin-fleck-at, line 54: Is this a reasonable default? I'm just wondering cause intuitively I'd say no because it is an optional service and not implementing an optional service should not stop something from working. However, I saw that in the client code we only ever send the action if we already know that the edge type hint is dynamic so we have a guard on that side. I'm not totally sure I think that should be sufficient, what do you think?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/typehints/RequestCheckEdgeTargetActionHandler.java
```diff
@@ -0,0 +1,75 @@
...
+      }
+      boolean hasDynamicHint = diagramConfiguration.getEdgeTypeHints().stream()
+         .filter(hint -> hint.getElementTypeId().equals(action.getEdgeType()) && hint.isDynamic()).findAny()
+         .isPresent();
+      if (!hasDynamicHint) {
+         return false;
```
martin-fleck-at, line 60: Similar question: Is this a good default value?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/typehints/RequestTypeHintsActionHandler.java
```diff
@@ -13,15 +13,21 @@
...
 import org.eclipse.glsp.server.actions.AbstractActionHandler;
+import org.eclipse.glsp.server.actions.Action;
+import org.eclipse.glsp.server.diagram.DiagramConfiguration;
 
 import com.google.inject.Inject;
 
+/**
+ * Default handler for {@link RequestTypeHintsAction}s.
+ * Queries the type hints provided by the {@link DiagramConfiguration}
+ *
```
martin-fleck-at, line 29:
```suggestion
 *
```

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/typehints/CheckEdgeTargetResultAction.java
```diff
@@ -0,0 +1,79 @@
...
+ ********************************************************************************/
+package org.eclipse.glsp.server.features.typehints;
+
+import java.util.Optional;
+
+import org.eclipse.glsp.server.actions.ResponseAction;
+
+/**
+ * Response Action for a {@link RequestCheckEdgeAction}. It provides
+ * a boolean indicating whether the requested element is a valid target
```
martin-fleck-at, line 24: the `requested element` is a bit confusing. I added some comments in the client change as well, I think the doc could be a bit clearer in what this action does and avoid the word `target` in the context of edges.

martin-fleck-at (changes requested):
Change looks good to me on a high level. I just have some minor concerns regarding documentation and default values. Thank you!

Btw, is there a plan to also adapt this in the Node server?

tortmayr:
>Btw, is there a plan to also adapt this in the Node server?
The PR for the node server will be openend once the client change is merged (because we need the protocol changes there)

martin-fleck-at (approved):
LGTM!

PR #218, opened 2023-10-11
GLSP-1127: Fix publish build

tortmayr:
The tycho-ant-run task for creating the composite update site seems to break with Java17. Since this task is just creating the update site the Java version is not really relevant so for now we switch back to 11. Update jenkins file to trigger Java21 compatibility when on master merges.


martin-fleck-at (approved):
LGTM!

PR #219, opened 2023-10-13
Bugfixes for 2.0

tortmayr:
Fix GModelDeleteOperationHandler by updating the collectDependents method.
- Dont add child elements of elements that should be deleted to the dependent set. They will be deleted indirectly anyways because the parent container is deleted.
- Correctly add incoming/outgoing edges to the dependet set instead of their targets/sources


martin-fleck-at (approved):
Works great, thank you!

PR #223, opened 2023-12-15
Fixes https://github.com/eclipse-glsp/glsp/issues/1199

dmm9:


martin-fleck-at (approved):
Thank you @dmm9! Change looks good to me!

PR #224, opened 2024-01-24
Fixes eclipse-glsp/glsp#1212

ivy-cst:
Always execute post update queue even if there is no handler for the UpdateModelAction


martin-fleck-at (approved):
Thank you @ivy-cst! Code looks good to me and the semantic change ("post update" even if there is no handler for the "update") is also fine with me. By default the client should register as an UpdateModelAction handler but I agree that we should clear the queue even if that is not the case.

CI fails as we are in the middle of a release so we need to update some other things before merging this.

ivy-cst:
> Thank you @ivy-cst! Code looks good to me and the semantic change ("post update" even if there is no handler for the "update") is also fine with me. By default the client should register as an UpdateModelAction handler but I agree that we should clear the queue even if that is not the case.
> 
> CI fails as we are in the middle of a release so we need to update some other things before merging this.

May be it is also because we use old clients at the moment which do not provide the `clientActionKinds`.
But at the moment I do not see where a `UpdateModelActionHandler` is created or registered?
But looks like I miss something here.

ivy-cst:
Thanks for the merge before release :+1:

PR #227, opened 2024-02-01
GLSP collaboration

hege088:
Initial implementation of the GLSP collaboration mode (see https://github.com/eclipse-glsp/glsp/issues/1279)


planger:
Thank you very much for this exciting feature! We'll review everything together once also the PRs for the VS Code integration and GLSP client are in. Can you please link them here so we know how to set up and test this feature?
Thank you!

PR #229, opened 2024-04-05
Extended workflow example by injecting non-default state

Sakrafux:
@planger As was discussed, the workflow example on the server was extended with a custom implementation for `ModelState`, `ModelStorage` and `GModelFactory`. However, since the Workflow implementation and its handlers are heavily dependent on the source model already being a `GModel`. This makes the `GModelFactory` irrelevant (or even detrimental), and I therefore opted to split the default functionality from `GModelStorage` to better show how a custom implementation could look like.

This is sufficient imo, since I will most likely need to implement a small custom model anyway for the planned editor examples.


planger:
Thank you very much for the PR! I think this improves the situation already.

However, I believe it'd be even more useful, if we actually switch to a very simple JSON format as source model, also if this means we'd have to also adapt the operation handlers. In practice, the GModel is rarely used directly as source model, as most people would like to use a custom model format. So the main example should probably do the same.

The only issue in the Java-based server is how to provide undo/redo support. The Java-server only provides that out of the box for EMF-based models. Otherwise it would have to be implemented manually, which isn't rocket science but adds some extra overhead.

What do you think?

PR #231, opened 2024-05-27
Add Resizable with 'resizeLocations' to shape elements

martin-fleck-at:
#### What it does

- Use string for GResizeLocation
- Fix issue with issue severity not being of type string

#### Changelog


tortmayr (approved):
Looks good to me an works like a charm. Thanks!

PR #232, opened 2024-05-29
Configure fork and join nodes as not resizable

tortmayr:
#### What it does

Update type hints configuration for workflow example to make fork and join nodes not resizable => we have test cases for both resizable and non-resizable elements in the diagram#### How to test

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #233, opened 2024-06-11
Minor improvements

tortmayr:
#### What it does
- Ensure that the ActionDispatchers postUpdateQueue is also cleared after  the inital `SetModelAction`
-  Change visibility of private members to protected for better extensibility
- Ensuare that Request and Response actions are initialized with an empty string as request/response id. This is required because otherwise typechecks on the client might fail

#### Changelog


planger (approved):

PR #238, opened 2024-06-24
Fix workflow documentation navigation example

tortmayr:
#### What it does

Update the NodeDocumentationNavigationTargetProvider to not rely on the tasks name which is currently unreliable (see https://github.com/eclipse-glsp/glsp/issues/1351) and instead relies on the `text` property of the tasks label.

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #241, opened 2024-08-09
Collab advanced

MatthiasHofstaetter:
#### Changelog


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/gmodel/AbstractGModelCreateNodeOperationHandler.java
```diff
@@ -59,7 +60,10 @@ public void executeOperation(final CreateNodeOperation operation) {
          .map(location -> LayoutUtil.getRelativeLocation(location, container));
       GModelElement element = createNode(relativeLocation, operation.getArgs());
       container.getChildren().add(element);
-      actionDispatcher.dispatchAfterNextUpdate(new SelectAction(), new SelectAction(List.of(element.getId())));
+      actionDispatcher.dispatchAfterNextUpdate(
```
tortmayr, line 63: Having to manually add the Subclient id to actions that should be dispatched after the next update is not ideal.
Wouldn't it be better to rework the `dispatchAfterNextUpdate` mechanism in the action dispatcher directly?
We should store rework the postUpdateQueue to a Map<string,List<Action>>. Whenever the `dispatchAfternextUpdate` method is invoked we then store the action in the map entry for the corresponding subclient id of the action.
When we clear the postUpdateQueue (`dispatchPostUpdateQueue`) we should only remove the actions that have the same subclientid as the corresponding updateModelAction.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/Action.java
```diff
@@ -51,6 +62,8 @@ public Action(final String kind) {
 
    public void setReceivedFromClient(final boolean receivedFromClient) { this.receivedFromClient = receivedFromClient; }
 
+   public String getSubclientId() { return subclientId; }
```
tortmayr, line 65: I think it would be good to already return the FALLBACK_SUBCLIENT_ID here if the id is null.
This would ensure that the subClientId is always defined and we don't run into potential null-pointers in the future.

PR #244, opened 2024-08-27
GLSP-1380: Remove dependency to Guava

### LINKED ISSUE #1380: Check Guava dependency in Java Server
Currently the Guava version is restricted to `[30.1.0,33.0.0)`. We should check why we have the upper bound of 33.0.0 as adopters may want to use newer versions. As part of this task we may also investigate whether we actually need Guava at all at this point.

It seems that Guava is mostly used for collection creation, precondition checking and the BiMap that we use in the indices. While the BiMap may be a good reason to keep the dependency, collection creation can be done differently and precondition checking is a simple Boolean check with an exception on false, so they might not be.

I believe the main problem may be the re-export and the upper bound so we should re-think that as part of this ticket.

tortmayr:
#### What it does
Removes the dependency to Google Guava. We mostly used collection creation which can be done with native Java now almost as conveniently. Other than that we only used the BiMap for source model indexing in EMF. While Guavas BIMaps are powerful most of their functionality is not required for our usecases therefore the useage got replaced with a custom BiIndex implementation. 

 - Replaces usage of Lists.newArrayList etc. with Java native calls
 -  Guava Preconditions were only used in two places -> replaced with in method exception handling
- Introduce a `BiIndex` helper class that serves as a replacement for Guavas BiMap in source model indexing
- Add test cases for the new BiIndex
- Update target platforms

Fixes https://github.com/eclipse-glsp/glsp/issues/1380

Contributed on behalf of AxonIvy AG

#### Changelog


examples/org.eclipse.glsp.example.workflow/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,156 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
martin-fleck-at, line 1: Why did we need to add this file, could you please elaborate on that?

plugins/org.eclipse.glsp.layout/src/org/eclipse/glsp/layout/ElkLayoutEngine.java
```diff
@@ -271,7 +272,12 @@ protected void handleClientLayout(final GBoundsAware element, final GLayouting s
     * Resolve cross-references in the ELK graph.
     */
    protected void resolveReferences(final LayoutContext context) {
-      Map<String, ElkConnectableShape> id2NodeMap = Maps.newHashMapWithExpectedSize(context.shapeMap.size());
+      // Per default a map will resize when 75% of its capacticy is filled. By dividing the expected size with the load
+      // factor
```
martin-fleck-at, line 276: unexpected newline here with only a single word in a line

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/utils/BiIndexTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class BiIndexTest {
```
martin-fleck-at, line 26: Always good to add tests, thank you!

martin-fleck-at (approved):
Code looks good to me, in general. Everything still seems to work as expected. I just a minor nitpick and I'm wondering about the `dependency-reduced-pom.xml`. Could you please elaborate?

martin-fleck-at:
This should probably mentioned in the changelog or do you think it won't be necessary?

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/utils/BiIndexTest.java
tortmayr: Agreed. In this case its especially import because we don't have an in-repo example to test the EMF integration.
Even with this test case we should probably still test the changes with one of the emf-project-templates in the example repo once the nightly version is available

tortmayr:
> This should probably mentioned in the changelog or do you think it won't be necessary?

Yes, you are right. I forgot to tick the checkboxes. Should definitly be added to the changelog

martin-fleck-at (approved):
LGTM!

PR #247, opened 2024-10-18
GLSP-1408: Make live validation asynchronous

martin-fleck-at:
#### What it does

- Introduce asynchronous live validation behavior
- Send status message

https://github.com/eclipse-glsp/glsp/issues/1408

#### How to test

- Any workflow example
- Add breakpoint to the validation so that it takes longer and you can see the message
- See that while breakpoint is active you can perform other operations

#### Changelog


martin-fleck-at:
Note that I opted for the status message as it is less intrusive as the progress message. However, there can only ever be a single status message at a time so it might not be the right choice in all cases. I'm happy to discuss this.

@ivy-lli FYI as well

ivy-lli (approved):
Looks good to me 👍 Thanks a lot

martin-fleck-at:
@tortmayr I'll merge this for now and if you do not have any objections, I'll double check the Node server to behave the same.

PR #249, opened 2024-12-04
Update target platform to r2024-09 

tortmayr:
#### What it does

- Update target platforms.
   -New default target Eclipse 2024-09
  - New minimum baseline Eclipse 2023-12
  - Switch to jarkarta variant of lsp4j websocket
  - Switch to Jetty 12

Builds on top of: https://github.com/eclipse-glsp/glsp-server/pull/24

Changes done on top:
- Remove outdated targets
- Adjust version ranges
- Create a new target definition for the minimum supported version (2023-12)
- Restore explicit reexport of certain packages to maintain backwards compatibility
- Fix and align m2 build
- Add possibility to change the target platform of the p2 build with the p2.target property

Part of https://github.com/eclipse-glsp/glsp/issues/1397

#### How to test

Building and running the example via CLI should work as expected
Test the following:
P2:
- mvn clean verify -Pp2
- mvn clean  verifiy -Dp2.target=r2023-12 (to build against the minimum p2 target)
M2:
- mvn clean verify -Pm2
Build the example fatjar
- mvn clean verify -Pm2 -Pfatjar
Start the example jar
- java -jar examples/org.eclipse.glsp.example.workflow/target/org.eclipse.glsp.example.workflow-2.3.0-SNAPSHOT-glsp.jar -p 80
81 --websocket
(and open the example diagram.html of the standalone client example)

Eclipse version used for development: 2024-06 Modelling Tools
Also tested with 2023-12 Modellingtools

#### Changelog


releng/org.eclipse.glsp.repository/category.xml
```diff
@@ -67,6 +55,7 @@
       </description>
    </category-def>
    <category-def name="org.eclipse.glsp.dependencies" label="External Dependencies"/>
-   <repository-reference location="mvn:org.eclipse.jetty:jetty-p2:10.0.13:zip:p2site" enabled="true" />
-   <repository-reference location="https://download.eclipse.org/elk/updates/releases/0.8.1/" enabled="true" />
+   <repository-reference location="https://download.eclipse.org/elk/updates/releases/0.9.1/" name="ELK 0.9.1" enabled="true" />
```
martin-fleck-at, line 58: Should we also update the ELK version in the MANIFEST files or keep it at 0.8.1? Seems like we only build with 0.9.1 anymore.

martin-fleck-at (approved):
Change looks good to me, thank you very much, Tobias! I built and ran the server from maven and directly from Eclipse. It all seemed to work as expected. I just have a minor comment but nothing too important.

releng/org.eclipse.glsp.repository/category.xml
tortmayr: I don't think its necessary. Raising the minimum version without an explicit need might unnecessarily break adopters products. If you want we could adapt the target for the minimum version to use elk 0.8.1.
Then we are on the safe side an know that it works with both versions

martin-fleck-at: Perfect, just wanted to make sure, thank you!

tortmayr:
@martin-fleck-at The target for r2023-12 has been updated to use elk 0.8.1

PR #255, opened 2025-04-17
GLSP-1509: Fix maven build

tortmayr:
#### What it does
- Use correct qualified snapshot version
- Add explicit TypeAdapter for ELists to Graph JSON configurator (required due to a breaking change in GSON 2.13)

P2 build uses 2.11 but m2build can pull in a higher version. 
Nevertheless the fix is straight forward and prevents upgrade pains in the future so its preferrable over pinning
GSON to a version < 2.13.0

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #257, opened 2025-04-18
GH-1510: Add null check to ElkLayoutEngine

tortmayr:
#### What it does

Add null check for parent position to `getOffset` computation.
 Fixes eclipse-glsp/glsp/issues/1510

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #262, opened 2025-09-08
Fix m2 build

tortmayr:
#### What it does
Add name attribute to all pom files to align with the requirements for publishing to maven central

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #283, opened 2026-04-27
feat: add request/response support to server ActionDispatcher

martin-fleck-at:
#### What it does

Enable the server to send requests to the client and await responses, and to issue requests handled locally by server-side handlers.

- Add `request()` and `requestUntil()` to `ActionDispatcher`
- Intercept matching responses in `dispatch()` to resolve pending requests
- Translate `RejectAction` responses to exceptional completion
- Add stale-marker grace cleanup for late-response filtering
- Make `RequestAction.requestId` mutable; add `RequestAction.timeout`
- Add `ResponseAction.hasValidResponseId` utility
- Tighten `ClientActionForwarder` to skip ResponseActions without id
- Handle inbound `RequestAction` in `DefaultGLSPServer.process()`
- Make `dispatchAll` sequential while keeping the list-of-futures API
- Add tests for request/response, timeouts, late responses, and dispose

Relates to https://github.com/eclipse-glsp/glsp/issues/607

#### Changelog

`dispatchAll` executes actions sequentially now and will fail on the first rejected action.

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/ActionDispatcher.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/ActionDispatcher.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019-2021 EclipseSource and others.
+ * Copyright (c) 2019-2026 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -15,9 +15,9 @@
  ********************************************************************************/
 package org.eclipse.glsp.server.actions;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
-import java.util.stream.Collectors;
 
 import org.eclipse.glsp.server.disposable.IDisposable;
 import org.eclipse.glsp.server.features.core.model.UpdateModelAction;
@@ -27,6 +27,12 @@ import org.eclipse.glsp.server.features.core.model.UpdateModelAction;
  * handlers.
  */
 public interface ActionDispatcher extends IDisposable {
+
+   /**
+    * Default timeout in milliseconds applied by {@link #requestUntil(RequestAction)} when neither
+    * a timeout argument nor {@link RequestAction#getTimeout()} provides one.
+    */
+   long DEFAULT_REQUEST_TIMEOUT_MS = 2_000L;
 
    /**
     * <p>
@@ -40,15 +46,21 @@ public interface ActionDispatcher extends IDisposable {
    CompletableFuture<Void> dispatch(Action action);
 
    /**
-    * <p>
-    * Processes all given actions by dispatching to the corresponding handlers.
-    * </p>
+    * Dispatches the given actions sequentially. A failure short-circuits the batch; remaining
+    * actions are not dispatched and their futures complete with the upstream failure.
     *
     * @param actions Actions to dispatch
     * @return A list of {@link CompletableFuture CompletableFutures}; one for each dispatched action
     */
    default List<CompletableFuture<Void>> dispatchAll(final List<Action> actions) {
-      return actions.stream().map(action -> dispatch(action)).collect(Collectors.toList());
+      List<CompletableFuture<Void>> results = new ArrayList<>(actions.size());
+      CompletableFuture<Void> previous = CompletableFuture.completedFuture(null);
+      for (Action action : actions) {
+         CompletableFuture<Void> current = previous.thenCompose(v -> dispatch(action));
+         results.add(current);
+         previous = current;
+      }
+      return results;
    }
 
    /**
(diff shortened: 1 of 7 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/ActionDispatcher.java
```diff
@@ -40,15 +46,21 @@ public interface ActionDispatcher extends IDisposable {
...
-    * </p>
+    * Dispatches the given actions sequentially. A failure short-circuits the batch; remaining
+    * actions are not dispatched and their futures complete with the upstream failure.
     *
     * @param actions Actions to dispatch
     * @return A list of {@link CompletableFuture CompletableFutures}; one for each dispatched action
     */
    default List<CompletableFuture<Void>> dispatchAll(final List<Action> actions) {
-      return actions.stream().map(action -> dispatch(action)).collect(Collectors.toList());
+      List<CompletableFuture<Void>> results = new ArrayList<>(actions.size());
```
tortmayr, line 56: What is the reason for changing this behavior?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/RequestAction.java
```diff
@@ -33,4 +34,13 @@ public RequestAction(final String kind, final String requestId) {
    }
 
    public String getRequestId() { return requestId; }
+
+   public void setRequestId(final String requestId) { this.requestId = requestId; }
+
+   /**
+    * Maximum wait time in milliseconds for a response, or {@code null} for no timeout.
+    */
+   public Long getTimeout() { return timeout; }
```
tortmayr: Typically we use getters that return Optional.ofNullable for optional action fields.

tortmayr: We than also can use primitive long for the field

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/ActionDispatcher.java
tortmayr: Technically a break because we no have failure short-circuits which previously was not the case

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/internal/actions/DefaultActionDispatcherTest.java
```diff
@@ -0,0 +1,542 @@
...
+      assertEquals(resp2, f2.get(AWAIT_MS, TimeUnit.MILLISECONDS));
+      assertFalse(f1.isDone());
+
+      TestResponse resp1 = new TestResponse(r1.getRequestId());
+      dispatcher.dispatch(resp1);
+      assertEquals(resp1, f1.get(AWAIT_MS, TimeUnit.MILLISECONDS));
+   }
+
+   @Test
+   void reentrantRequestFromHandlerCorrelatesResponse() throws Exception {
```
tortmayr, line 167: The test dispatches a response from a new thread inside a handler, then does future.get(AWAIT_MS, ...). This works but
    relies on the new thread running before AWAIT_MS. In pathological CI environments with very high load, this could
    theoretically flake. The CountDownLatch pattern used elsewhere in the test class would make this deterministic.

tortmayr (changes requested):

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/ActionDispatcher.java
martin-fleck-at: Basically this is the same behavior that we have in Node where we execute the actions sequentially and the first await that fails will break the chain. We can keep the original one but that means we might be executing actions in parallel if I'm not mistaken so actions cannot build upon each others result. But we might want to double check this.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/RequestAction.java
martin-fleck-at: Fair, I'll change it. We just need some way to specify "no timeout".

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/internal/actions/DefaultActionDispatcherTest.java
martin-fleck-at: Will update!

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/actions/ActionDispatcher.java
tortmayr: Ah I see what you mean. 
In the reentrant case this could call to parallel execution because all bypass the queue right?
In that case lets change it but we should mention the changed behavior in the changelog

martin-fleck-at: Yes, or a user might call it explicitly and our general approach is to execute actions sequentially. I already ticked that this PR should be mentioned in the changelog, I make a note in the PR description.

tortmayr (approved):
LGTM! 👍🏼

PR #295, opened 2026-08-25
GLSP-1727: Keep applying computed bounds when an entry cannot be applied

martin-fleck-at:
#### What it does

A single unresolvable element in a ComputedBoundsAction aborted the whole batch: the client had already measured every element, but one stale id threw and the server dropped the remaining bounds, alignments and routes along with the model update the client was waiting for. Mirrors the same fix in glsp-server-node.

- resolve elements through the index Optional instead of getOrThrow, so an unknown id yields an empty result
- report a route with fewer than two points as not applicable rather than as an error, a client may have nothing to report for an unmeasured edge yet
- log a skipped entry, routes at debug level because an edge the client has not finished routing yet is expected
- split the apply step into applyElementBounds, applyAlignments and applyRoutes
- leave applyRoutingPoints strict, an unknown id in a ChangeRoutingPointsOperation is a real error

Adjusting the apply step previously meant overriding executeAction, which also takes over the revision check and the model lock. ComputedBoundsActionHandler now dispatches through overridable per-kind methods, so an adopter can replace one kind and rebind the handler, as the workflow example does for other handlers.

LayoutUtil.applyRoute now returns Optional<GEdge>; noted in the changelog under potentially breaking changes.

Relates to eclipse-glsp/glsp#1727

#### Changelog

### DIFF
--- a/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ComputedBoundsActionHandler.java
+++ b/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ComputedBoundsActionHandler.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019-2023 EclipseSource and others.
+ * Copyright (c) 2019-2026 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -17,10 +17,14 @@ package org.eclipse.glsp.server.features.core.model;
 
 import java.util.List;
 
+import org.eclipse.glsp.graph.GModelIndex;
 import org.eclipse.glsp.graph.GModelRoot;
 import org.eclipse.glsp.server.actions.AbstractActionHandler;
 import org.eclipse.glsp.server.actions.Action;
 import org.eclipse.glsp.server.model.GModelState;
+import org.eclipse.glsp.server.types.ElementAndAlignment;
+import org.eclipse.glsp.server.types.ElementAndBounds;
+import org.eclipse.glsp.server.types.ElementAndRoutingPoints;
 import org.eclipse.glsp.server.utils.LayoutUtil;
 
 import com.google.inject.Inject;
@@ -44,11 +48,59 @@ public class ComputedBoundsActionHandler extends AbstractActionHandler<ComputedB
          GModelRoot model = modelState.getRoot();
          if (model != null && action.getRevision().isPresent()
             && action.getRevision().get().doubleValue() == model.getRevision()) {
-            LayoutUtil.applyBounds(model, action, modelState);
+            applyBounds(model, action);
             return submissionHandler.submitModelDirectly();
          }
       }
       return none();
+   }
+
+   /**
+    * Applies everything the client computed for the given model.
+    *
+    * <p>
+    * Override this, or one of the per-kind methods it delegates to, to adjust what is applied. Overriding
+    * {@link #executeAction(ComputedBoundsAction)} instead would also take over the revision check and the model lock.
+    * </p>
+    *
+    * @param root   The model root.
+    * @param action The computed bounds action.
+    */
+   protected void applyBounds(final GModelRoot root, final ComputedBoundsAction action) {
+      GModelIndex index = modelState.getIndex();
+      applyElementBounds(action.getBounds(), index);
+      applyAlignments(action.getAlignments(), index);
+      applyRoutes(action.getRoutes(), index);
+   }
+
+   /**
+    * Applies the computed bounds of several elements.
+    *
+    * @param allBounds The new bounds.
+    * @param index     The model index.
+    */
+   protected void applyElementBounds(final List<ElementAndBounds> allBounds, final GModelIndex index) {
+      LayoutUtil.applyElementBounds(allBounds, index);
+   }
+
+   /**
+    * Applies the computed alignments of several elements.
+    *
+    * @param alignments The new alignments.
+    * @param index      The model index.
+    */
+   protected void applyAlignments(final List<ElementAndAlignment> alignments, final GModelIndex index) {
+      LayoutUtil.applyAlignments(alignments, index);
+   }
+
+   /**
+    * Applies the computed routes.
+    *
+    * @param routes The new routes.
+    * @param index  The model index.
+    */
+   protected void applyRoutes(final List<ElementAndRoutingPoints> routes, final GModelIndex index) {
+      LayoutUtil.applyRoutes(routes, index);
    }
 
 }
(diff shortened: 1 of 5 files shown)


plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ComputedBoundsActionHandler.java
```diff
@@ -44,11 +48,59 @@ public List<Action> executeAction(final ComputedBoundsAction action) {
...
+    * <p>
+    * Override this, or one of the per-kind methods it delegates to, to adjust what is applied. Overriding
+    * {@link #executeAction(ComputedBoundsAction)} instead would also take over the revision check and the model lock.
+    * </p>
+    *
+    * @param root   The model root.
+    * @param action The computed bounds action.
+    */
+   protected void applyBounds(final GModelRoot root, final ComputedBoundsAction action) {
+      GModelIndex index = modelState.getIndex();
```
tortmayr, line 70: Whats the point of passing the root in here if we don't use it? Can be dropped

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/utils/LayoutUtilTest.java
```diff
@@ -0,0 +1,158 @@
...
+   void reportsAlignmentOfAnUnresolvableElementAsNotApplied() {
+      assertTrue(LayoutUtil.applyAlignment(alignment(UNKNOWN_ID), index).isEmpty());
+   }
+
+   @Test
+   void appliesTheReportedRoute() {
+      GPoint source = GraphUtil.point(0, 0);
+      GPoint target = GraphUtil.point(10, 10);
+
+      assertTrue(LayoutUtil.applyRoute(route(EDGE_ID, source, GraphUtil.point(5, 5), target), index).isPresent());
```
tortmayr: This only asserts `isPresent()`, so the part of `applyRoute` that actually got rewritten stays uncovered. Assert that `edge.getRoutingPoints()` holds just the middle point and that `KEY_EDGE_SOURCE_POINT`/`KEY_EDGE_TARGET_POINT` got source and target, otherwise swapping the two `remove` calls would still pass.

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/utils/LayoutUtilTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.junit.jupiter.api.Test;
+
+public class LayoutUtilTest {
+
+   private static final String NODE_ID = "node0";
+   private static final String EDGE_ID = "edge0";
+   private static final String UNKNOWN_ID = "does-not-exist";
+
+   /** Exposes a fixed index, the injected one is not available outside of Guice. */
+   private static class TestGModelState extends DefaultGModelState {
```
tortmayr: `DefaultGModelState.updateRoot(root)` already builds and stores the index ([master](https://github.com/eclipse-glsp/glsp-server/blob/9be5c03098457cf16d34cae801184768f0e90865/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/model/DefaultGModelState.java#L64-L72)), so `new DefaultGModelState()` + `updateRoot(graph)` does the job without a subclass. The comment is also off: `index` is a plain field defaulting to `GModelIndex.empty()`, nothing injects it.

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/features/core/model/ComputedBoundsActionHandlerTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ComputedBoundsActionHandlerTest {
+
+   private static final String NODE_ID = "node0";
+   private static final String EDGE_ID = "edge0";
+
+   /** Exposes a fixed index, the injected one is not available outside of Guice. */
+   private static class TestGModelState extends DefaultGModelState {
```
tortmayr: Same `TestGModelState` copy as in `LayoutUtilTest`. Either drop it in favour of `DefaultGModelState.updateRoot(graph)` or put it in one shared test helper.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/utils/LayoutUtil.java
```diff
@@ -45,35 +47,86 @@
...
     *
+    * <p>
+    * An entry that cannot be applied is skipped and logged rather than treated as an error.
+    * </p>
+    *
     * @param root       The model root.
     * @param action     The computed bounds action.
     * @param modelState The model state
     */
    public static void applyBounds(final GModelRoot root, final ComputedBoundsAction action,
```
tortmayr, line 70: This is no longer activly called. The computed bounds action handler has its own `applyBounds` method now. Maybe we should deprecate it?

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/features/core/model/ComputedBoundsActionHandler.java
martin-fleck-at: Kept it for now. The node server has the same signature with the same unused root, so dropping it only here would make the two drift apart. It is also a protected extension point, so an override can take the root from the parameter instead of going through modelState. If you prefer it gone, I would rather drop it in both repos.

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/utils/LayoutUtilTest.java
martin-fleck-at: Assertions added. The routing points now have to hold just the middle point, and source and target have to end up under their args keys.

One correction on the reasoning: swapping the two remove calls changes nothing. For [a,b,c] you get a and c either way, and [b] stays behind. So that swap would still pass, even with the new assertions. What they do catch is the two put keys being mixed up, and the endpoints not being removed from the routing points.

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/utils/LayoutUtilTest.java
martin-fleck-at: Correct on both points. The subclass is gone and the wrong comment with it. This test file was the only place using it, and only in the batch test, which now calls the per-kind methods directly and needs no model state at all.

tests/org.eclipse.glsp.server.test/src/org/eclipse/glsp/server/features/core/model/ComputedBoundsActionHandlerTest.java
martin-fleck-at: Gone here too. This one uses new DefaultGModelState() plus updateRoot(graph), as you suggested. No copy is left in either file, so a shared helper is not needed.

plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/utils/LayoutUtil.java
martin-fleck-at: Deprecated, pointing at the handler applyBounds. It is worth doing because this method calls the static per-kind methods, so it skips the overridable ones on the handler. Someone who overrides applyRoutes would not see their override used here.

The three static per-kind methods stay as they are, the handler still delegates to them. Also moved the one test that used this method over to those, so nothing calls the deprecated one anymore. Changelog entry added.

tortmayr (approved):

